PR #2, opened 2015-02-20
Initial work on footer directive

darwinlopez:
Initial work on footer directive.
In future versions: HTML template might be broken apart into its own HTML file to be loaded with templateUrl - unless considered small enough to be broken
Functional tests still to be defined on further discussions


src/js/directives/footers/rb-footer.js
```diff
@@ -0,0 +1,70 @@
+(function () {
+    'use strict';
+
+    /**
+    * @ngdoc module
```
nwhite89: when doing comments these should have an extra indent of a space
```js
/**
 *
 */
```

src/js/directives/footers/rb-footer.js
```diff
@@ -0,0 +1,70 @@
...
+            },
+            restrict: 'E',
+            replace: true,
+            template: getTemplate
+        };
+
+        // Might be broken apart - in following stories - into its own HTML file
+        // to be loaded with templateUrl.
+        function getTemplate (elem, attr) {
+            return '<footer class="Footer" role="contentinfo">' +
```
nwhite89: can we please use something like https://www.npmjs.com/package/gulp-ngtemplate I really don't want HTML within the JS even from the start, using ngTemplate it will package the HTML into JS to prevent extra resource requestes.

src/js/directives/footers/rb-footer.js
chrisvxd: addressed

darwinlopez: Done

src/js/directives/footers/rb-footer.js
darwinlopez: Done

PR #12, opened 2015-04-07
Create the Rockabox date time component

nwhite89:
Creates the Rockabox date time component ``<rb-date-time>`` allowing to specify:
- No time ``show-time="false"``
- Understated time ``understate-time="true"``

TP: https://rockabox.tpondemand.com/entity/8270

### DIFF
--- a/src/rb-date-time/rb-date-time-directive.js
+++ b/src/rb-date-time/rb-date-time-directive.js
@@ -13,9 +13,9 @@ define([
      *
      * @usage
      * <hljs lang="html">
-     *    <rb-date-time dateTime="1970-01-01T00:00"></rb-date-time>
-     *    <rb-date-time dateTime="1970-01-01T00:00" understateTime="true"></rb-date-time>
-     *    <rb-date-time dateTime="1970-01-01T00:00" understateTime="true" show-time="false"></rb-date-time>
+     *    <rb-date-time date-time="1970-01-01T00:00"></rb-date-time>
+     *    <rb-date-time date-time="1970-01-01T00:00" understate-time="true"></rb-date-time>
+     *    <rb-date-time date-time="1970-01-01T00:00" show-time="false"></rb-date-time>
      * </hljs>
      *
      * @ngInject
@@ -30,7 +30,6 @@ define([
             },
             restrict: 'E',
             replace: true,
-            transclude: true,
             template: template
         };
     }
--- a/src/rb-date-time/rb-date-time-directive.js
+++ b/src/rb-date-time/rb-date-time-directive.js
@@ -1,0 +1,39 @@
+define([
+    'html!./rb-date-time.tpl.html'
+], function (template) {
+
+    /**
+     * @ngdoc directive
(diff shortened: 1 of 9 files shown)


src/rb-date-time/rb-date-time-directive.js
```diff
@@ -0,0 +1,39 @@
...
+     *
+     * @restrict E
+     *
+     * @description
+     *
+     * @usage
+     * <hljs lang="html">
+     *    <rb-date-time dateTime="1970-01-01T00:00"></rb-date-time>
+     *    <rb-date-time dateTime="1970-01-01T00:00" understateTime="true"></rb-date-time>
+     *    <rb-date-time dateTime="1970-01-01T00:00" understateTime="true" show-time="false"></rb-date-time>
```
chrisvxd: Why the mixture of camel-case and hyphenated?

src/rb-date-time/rb-date-time.tpl.html
```diff
@@ -0,0 +1,6 @@
+<time
+    class="DateTime"
+    ng-class="understateTime ? 'DateTime--understateTime' : ''"
```
chrisvxd: Wouldn't this be tidier using an object for the expression? `{DateTime--understateTime: understateTime}`

src/rb-date-time/rb-date-time-directive.js
```diff
@@ -0,0 +1,39 @@
...
+
+        return {
+            scope: {
+                dateTime: '@',
+                understateTime: '@',
+                showTime: '@'
+            },
+            restrict: 'E',
+            replace: true,
+            transclude: true,
```
chrisvxd: Transclusion isn't needed here

nwhite89: true tbh bit of copy pasta lol

src/rb-date-time/rb-date-time.tpl.html
nwhite89: that expression wouldn't work :confused: ?

src/rb-date-time/rb-date-time-directive.js
nwhite89: shouldn't be will change

src/rb-date-time/rb-date-time.tpl.html
chrisvxd: Angular processes objects like that for ng-class...

`2. If the expression evaluates to an object, then for each key-value pair of the object with a truthy value the corresponding key is used as a class name.`

https://docs.angularjs.org/api/ng/directive/ngClass

nwhite89: hmm ok will give it a go

src/rb-date-time/rb-date-time-directive.js
nwhite89: good to note it's in the generator currently we may want to remove that for now? @chrisvxd

PR #13, opened 2015-04-08
Currency directive

nwhite89:
- ``<rb-currency>`` currently you can only use ``$`` 
- Will display 18291.19 as ``$18,291.19``
- Doesn't at cents when ``.00``

TP: https://rockabox.tpondemand.com/entity/8271


src/rb-date-time/demo/demo-state.js
```diff
@@ -6,7 +6,7 @@ define([
     function demoState ($stateProvider, $urlRouterProvider) {
         $stateProvider.state('rb-date-time', {
             url: '/rb-date-time',
-            controller: 'demo-ctrl as demoCtrl',
+            controller: 'demo-date-time-ctrl as demoCtrl',
```
chrisvxd: typo

src/rb-date-time/demo/index.js
```diff
@@ -8,7 +8,7 @@ define([
             rbDateTime.name
         ])
         .config(demoState)
-        .controller('demo-ctrl', demoCtrl);
+        .controller('demo-date-time-ctrl', demoCtrl);
```
chrisvxd: typo

chrisvxd:
I think the point of this directive is to take any numerical value, and show it as currency, with a configurable number of decimal places. This implementation requires us to define the decimal places in the controller each time. Here's how I would probably expect this to work:

`<rb-currency amount="1000" decimal-places="2" />` renders to `$1,000.00`
`<rb-currency amount="1000" decimal-places="0" />` renders to `$1,000`
`<rb-currency amount="1000.5122" decimal-places="2" />` renders to `$1,000.51`

Obviously the story says only show 2 decimal places, so that can be hard-coded

src/rb-date-time/demo/index.js
nwhite89: no this is correct

chrisvxd: @nwhite89 this is the currency control, not datetime

chrisvxd: + it only needs to be unique in this module

chrisvxd: Oh wait, this is in the date-time component. Ignore this.

PR #15, opened 2015-04-08
Initial work on button component

darwinlopez:
Initial work on button component

### DIFF
--- a/src/rb-button/demo/demo-state.js
+++ b/src/rb-button/demo/demo-state.js
@@ -6,7 +6,7 @@ define([
     function demoState ($stateProvider, $urlRouterProvider) {
         $stateProvider.state('rb-button', {
             url: '/rb-button',
-            controller: 'demo-ctrl as demoCtrl',
+            controller: 'demo-button-ctrl as demoCtrl',
             template: template
         });
     }
--- a/src/rb-button/demo/demo-state.js
+++ b/src/rb-button/demo/demo-state.js
@@ -1,0 +1,15 @@
+define([
+    'html!./demo.tpl.html'
+], function (template) {
+
+    // @ngInject
+    function demoState ($stateProvider, $urlRouterProvider) {
+        $stateProvider.state('rb-button', {
+            url: '/rb-button',
+            controller: 'demo-ctrl as demoCtrl',
+            template: template
+        });
+    }
+
+    return demoState;
+});
--- a/src/rb-button/demo/demo.tpl.html
+++ b/src/rb-button/demo/demo.tpl.html
@@ -1,0 +1,8 @@
+<rb-button type="button" state="danger">Button</rb-button>
+<rb-button type="submit" state="warning" block="block">Submit</rb-button>
+<rb-button type="button">Button</rb-button>
+<rb-button state="positive" href="http://www.xe.com" size="small">Link</rb-button>
+<rb-button href="http://www.xe.com" size="large">Link</rb-button>
+<rb-button type="submit" state="danger" block="block" size="large">Submit</rb-button>
+<rb-button type="button" state="positive" size="small" ng-disabled="true">Button</rb-button>
+<rb-button type="button" state="warning" size="large" ng-disabled="true" ng-href="http://www.xe.com">Link</rb-button>
--- a/src/rb-button/demo/demo.tpl.html
+++ b/src/rb-button/demo/demo.tpl.html
@@ -1,8 +1,9 @@
+<rb-button>Simple</rb-button>
 <rb-button type="button" state="danger">Button</rb-button>
-<rb-button type="submit" state="warning" block="block">Submit</rb-button>
+<rb-button type="submit" state="warning" block="yes">Submit</rb-button>
 <rb-button type="button">Button</rb-button>
 <rb-button state="positive" href="http://www.xe.com" size="small">Link</rb-button>
 <rb-button href="http://www.xe.com" size="large">Link</rb-button>
-<rb-button type="submit" state="danger" block="block" size="large">Submit</rb-button>
+<rb-button type="submit" state="danger" block="yes" size="large">Submit</rb-button>
 <rb-button type="button" state="positive" size="small" ng-disabled="true">Button</rb-button>
 <rb-button type="button" state="warning" size="large" ng-disabled="true" ng-href="http://www.xe.com">Link</rb-button>
--- a/src/rb-button/demo/index.js
+++ b/src/rb-button/demo/index.js
@@ -1,0 +1,14 @@
+define([
+    '../index.js',
+    './demo-ctrl',
+    './demo-state'
+], function (rbButton, demoCtrl, demoState) {
+    var demo = angular
+        .module('rb-button-demo', [
+            rbButton.name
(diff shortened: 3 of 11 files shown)


src/rb-button/demo/demo-state.js
```diff
@@ -0,0 +1,15 @@
+define([
+    'html!./demo.tpl.html'
+], function (template) {
+
+    // @ngInject
+    function demoState ($stateProvider, $urlRouterProvider) {
+        $stateProvider.state('rb-button', {
+            url: '/rb-button',
+            controller: 'demo-ctrl as demoCtrl',
```
nwhite89: can you rename ``demo-ctrl`` to something like ``demo-rb-button-ctrl`` have addressed this issue caused by the generator

src/rb-button/demo/index.js
```diff
@@ -0,0 +1,14 @@
...
+    '../index.js',
+    './demo-ctrl',
+    './demo-state'
+], function (rbButton, demoCtrl, demoState) {
+    var demo = angular
+        .module('rb-button-demo', [
+            rbButton.name
+        ])
+        .config(demoState)
+        .controller('demo-ctrl', demoCtrl);
```
nwhite89: will need to also be amended from previous comment ``demo-db-button-ctrl``

src/rb-button/demo/demo.tpl.html
```diff
@@ -0,0 +1,8 @@
+<rb-button type="button" state="danger">Button</rb-button>
+<rb-button type="submit" state="warning" block="block">Submit</rb-button>
+<rb-button type="button">Button</rb-button>
+<rb-button state="positive" href="http://www.xe.com" size="small">Link</rb-button>
+<rb-button href="http://www.xe.com" size="large">Link</rb-button>
+<rb-button type="submit" state="danger" block="block" size="large">Submit</rb-button>
```
chrisvxd: can `block` be a boolean instead?

src/rb-button/rb-button-directive.js
```diff
@@ -0,0 +1,127 @@
...
+     * @name rbButton
+     * @module directives.buttons.rb-button
+     *
+     * @restrict E
+     *
+     * @description
+     * `<rb-button>` is a plain button directive
+     *
+     * If you supply a `type` attribute with `submit` as value, it will become
+     * a submit button `<input type="submit"/>`. If you supply a `href`, `ng-href`
```
nwhite89: @ohskylab removed the need of having a ``<button type="submit"/>`` / ``<input type="submit" />`` could we remove please

src/rb-button/rb-button-directive.js
```diff
@@ -0,0 +1,127 @@
...
+        }
+
+        function isStatePresent (attr) {
+            return angular.isDefined(attr.state);
+        }
+
+        function getTemplate (elem, attr) {
+            if (isAnchor(attr)) {
+                return linkTemplate;
+            } else if (isSubmit(attr)) {
```
nwhite89: this will then be removed ?

src/rb-button/rb-button-submit.tpl.html
```diff
@@ -0,0 +1 @@
+<button type="submit" class="Button" ng-transclude></button>
```
nwhite89: again this will then be removed?

src/rb-button/demo/demo.tpl.html
nwhite89: sounds like it should be ? will also allow putting it straight into the template rather than having JS decide on the class?

src/rb-button/rb-button-directive.js
```diff
@@ -0,0 +1,127 @@
...
+            // Defaults to default size when no
+            // size is supplied
+            if (isSizePresent(attr)) {
+                elem.addClass('Button--' + attr.size);
+            }
+
+            // Defaults to blue colour when no
+            // state is supplied (Default state)
+            if (isStatePresent(attr)) {
+                elem.addClass('Button--' + attr.state);
```
nwhite89: could this not use ng-class? something like ``{ 'Button--' + attr.state: attr.state}``

nwhite89:
I would of imagined that it be possible to directly call a function ? or emit something on a button click? we're going to need this for something like publishing. Although I guess we can add that later when needed

src/rb-button/demo/demo.tpl.html
```diff
@@ -0,0 +1,8 @@
...
+<rb-button type="button" state="positive" size="small" ng-disabled="true">Button</rb-button>
+<rb-button type="button" state="warning" size="large" ng-disabled="true" ng-href="http://www.xe.com">Link</rb-button>
```
chrisvxd: I think the use of `state` could be confused with the state machine

src/rb-button/rb-button-directive.js
```diff
@@ -0,0 +1,127 @@
...
+            }
+
+            if (isBlock(attr)) {
+                elem.addClass('Button--block');
+            }
```
chrisvxd: Isn't this (and above 2 code blocks) what `ng-class` is for?

src/rb-button/demo/demo.tpl.html
darwinlopez: You can call a function, you can use attributes directives in any of the button... that's the first thing I'm checking on the tests... so you can use ng-<something> whenever you want

chrisvxd: I mean semantically more than programatically. It looks confusing!

src/rb-button/rb-button-directive.js
chrisvxd: Ditto (see below)

src/rb-button/demo/demo.tpl.html
nwhite89: state is what @ohskylab has called them state of button right?

src/rb-button/rb-button-directive.js
nwhite89: yeah @darwinlopez I'd rather us use something like ng-class rather than having to have functions to do the same sort of thing unless there is a specific reason for doing it?

darwinlopez: How would you use ng-class here ? I can't see how it would help

nwhite89: we're doing the same in date-time

```
ng-class="{
        'DateTime--understateTime': understateTime
    }"
```

src/rb-button/rb-button-directive.js
```diff
@@ -0,0 +1,127 @@
...
+            return angular.isDefined(attr.type) &&
+                angular.equals(angular.lowercase(attr.type), 'submit');
+        }
+
+        function isBlock (attr) {
+            return angular.isDefined(attr.block) &&
+                angular.equals(angular.lowercase(attr.block), 'block');
+        }
+
+        function isSizePresent (attr) {
```
nwhite89: can you stop using functions unnecessarily this for example and ``isStatePresent`` are doing the same and is purely a wrapper around isDefined

nwhite89:
MoG

PR #16, opened 2015-04-09
Adds a Loading component

nwhite89:
Creates the loading component, allowing it to be passed a value to describe what is "Loading"

TP: https://rockabox.tpondemand.com/entity/8643

### DIFF
--- a/src/rb-loading/rb-loading-directive.js
+++ b/src/rb-loading/rb-loading-directive.js
@@ -1,0 +1,34 @@
+define([
+    'html!./rb-loading.tpl.html'
+], function (template) {
+
+    /**
+     * @ngdoc directive
+     * @name rbLoading
+     * @module rb-loading
+     *
+     * @restrict E
+     *
+     * @description
+     *
+     * @usage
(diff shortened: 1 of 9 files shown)


src/rb-loading/rb-loading-directive.js
```diff
@@ -0,0 +1,34 @@
...
+     *    <rb-loading item="People"></rb-loading>
+     * </hljs>
+     *
+     * @ngInject
+     */
+    function rbLoadingDirective () {
+
+        return {
+            scope: {
+                item: '@'
```
chrisvxd: Would prefer this to be "message"

src/rb-loading/rb-loading.tpl.html
```diff
@@ -0,0 +1,8 @@
+<div class="Loading">
+    <div class="Loading-spinner"></div>
+    <div class="Loading-message">
+        <p>
+            Loading {{ ::item }}
```
chrisvxd: I think `Loading` should be extrapolated out into `message` (see above)

PR #20, opened 2015-04-14
Angularify badge component.

k0nG:
TP Story:
https://rockabox.tpondemand.com/entity/8475

Task:
https://rockabox.tpondemand.com/entity/8284

Angularised/Angularfied badge component. Depends on https://github.com/rockabox/rbx_style_guide/pull/151

### DIFF
--- a/src/rb-badge/rb-badge-directive.js
+++ b/src/rb-badge/rb-badge-directive.js
@@ -47,25 +47,29 @@ define([
 
         return {
             scope: {
-                collapsed: '@'
+                collapsed: '@',
+                state: '@'
             },
             restrict: 'E',
             replace: true,
             link: link,
-            template: template
+            template: template,
+            controller: controller
+        };
+    }
+
+    function controller ($scope) {
+
+        // Allow template to check a state is valid.
+        $scope.validState = function (state) {
+            return state in STATE_DETAILS;
         };
     }
 
     function link (scope, elem, attr) {
 
-        // Set state class and details string
-        // Only set states defined in STATE_DETAILS
-        if (angular.isDefined(attr.state) && attr.state in STATE_DETAILS) {
-            elem.addClass('Badge--' + attr.state);
-            scope.details = STATE_DETAILS[attr.state];
-        } else {
-            elem.html('');
-        }
+        // Pass details lookup to scope.
+        scope.details = STATE_DETAILS;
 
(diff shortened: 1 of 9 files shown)


src/rb-badge/rb-badge-directive.js
```diff
@@ -0,0 +1,81 @@
...
+
+        // Set body text for warning state. Defaults to &nbsp; in template
+        if (angular.isDefined(attr.body) && attr.state == 'warning') {
+            scope.body = attr.body + ' !';
+        }
+
+        // Set is-collapsed class
+        if (angular.isDefined(attr.collapsed)) {
+            elem.addClass('is-collapsed');
+        }
```
chrisvxd: I prefer use of ng-class over assigning classes in js

k0nG: No that makes sense. So pass it through on the scope then use ng-if. I'll update.

nwhite89: should be able to do
```
ng-class="{
  'is-collapsed': collapsed
}"
```

take a look here https://github.com/rockabox/rb-table/blob/v1/src/directives/table/table.tpl.html#L19

src/rb-badge/rb-badge-directive.js
```diff
@@ -0,0 +1,80 @@
...
+     *    <rb-badge state="warning" body="6!"></rb-badge>
+     *    <rb-badge state="statusComplete"></rb-badge>
+     * </hljs>
+     *
+     * @ngInject
+     */
+    function rbBadgeDirective () {
+
+        return {
+            scope: {},
```
chrisvxd: You should use isolated scope for the attrs so you can stop handling vars via `attr` in `link`

src/rb-badge/rb-badge-directive.js
```diff
@@ -0,0 +1,80 @@
...
+
+        // Set body text for warning state. Defaults to &nbsp; in template
+        if (angular.isDefined(attr.body) && attr.state == 'warning') {
+            scope.body = attr.body + ' !';
+        }
+
+        // Set is-collapsed class
+        if (angular.isDefined(attr.collapsed)) {
+            scope.collapsed = true;
+        }
```
chrisvxd: This method will go if we use isolated scopes

PR #25, opened 2015-04-15
ng-class level unit tests for text-control

chrisvxd:
Should have been testing this before

TP https://rockabox.tpondemand.com/entity/8772


k0nG:
MOG

PR #27, opened 2015-04-15
Have css with the date time component

nwhite89:
Also sets up the build process to use post-css

TP: https://rockabox.tpondemand.com/entity/8818


k0nG:
OK by me. But think @chrisvxd should look at it as I don't know much about webpack..

PR #32, opened 2015-04-16
rb-nav-bar

nwhite89:
- Move's over the ``rb-nav-bar`` css
- Update ui component's to build with all of rbx_style_guide's post-css dependencies

TP: https://rockabox.tpondemand.com/entity/8824


webpack.base.config.js
```diff
@@ -57,12 +58,24 @@ config = {
     },
     postcss: [
         webpackPostcssTools.prependTildesToImports,
-        require('autoprefixer-core')({
-            remove: false,
-            cascade: false
+        require('postcss-import')({
+            path: [
+                path.join('vendor/rbx_style_guide/src/assets/styles')
```
chrisvxd: this `path.join` is not necessary

nwhite89: meant to have __dirname, 'vendor/' will update

PR #38, opened 2015-04-16
Fixed broken login form dependencies

chrisvxd:
Accidentally set these up wrong so couldn't integrate into campaign management.

No tp.


k0nG:
MOG

PR #49, opened 2015-04-20
rb-text-control readonly support

chrisvxd:
you can use the `is-readonly` attr on the directive

TP https://rockabox.tpondemand.com/entity/8901


nwhite89:
MoG :gree

PR #50, opened 2015-04-21
Initial work on warning messages

darwinlopez:
Initial work on warning messages

In a live environment the click and dismiss function should live in a warning service and be exposed to any controller(s) you want, same with the warning list.

### DIFF
--- a/dev/app.js
+++ b/dev/app.js
@@ -39,10 +39,6 @@ define([
     rbWarningMessagesDemo,
     css
 ) {
-    uiRouter, template, rbBadgeDemo, rbButtonDemo, rbDateTimeDemo, rbFooterDemo, rbHeaderDemo, rbLoadingDemo,
-    rbLoginFormDemo, rbMainDemo, rbPageHeaderDemo, rbPageTitleDemo, rbRatioDemo, rbSiteDemo, rbSystemMessage,
-    rbTextControlDemo, rbWarningMessagesDemo, css
-    ) {
     // @ngInject
     angular
         .module('ui-demo', [
--- a/dev/app.js
+++ b/dev/app.js
@@ -16,6 +16,7 @@ define([
     'components/rb-site/demo',
     'components/rb-system-message/demo',
     'components/rb-text-control/demo',
+    'components/rb-warning-messages/demo',
     'rbx_style_guide'
 ], function (
     uiRouter,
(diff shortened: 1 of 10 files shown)


nwhite89:
@darwinlopez could you explain what the idea is, in a live environment do you simply pass a resource? or do you need to pass a custom function per controller that uses it?

dev/app.js
```diff
@@ -62,5 +64,16 @@ define([
...
+        .value('warningStates', {
+            // A map of all available WARNING_IDENTIFIER: UI_ROUTER_STATE
+            // (key:value) pairs
+            // Everytime a user clicks an issue link on any warning message,
+            // will be a redirection to a UI-Router state.
+            // UI-Router states will be determined by this key:value map
+            // with key: WARNING_IDENTIFIER
+            // STATES FOR DEMO PURPOSES ONLY
+            'PARTIAL_TRACKING': 'tracking',
+            'OTHER_STATE': 'other_state'
```
chrisvxd: What do you think about moving this to `src/rb-warning-messages/demo/index.js`?

src/rb-warning-messages/demo/demo-ctrl.js
```diff
@@ -0,0 +1,112 @@
...
+
+        // FOR DEMO PURPOSES ONLY
+        this.dismissy = function (id) {
+
+            console.log('Dismissing the warning with id: ' + id);
+            console.log('... and after waiting 2.5 seconds...');
+            // REMOVES THE WARNING FROM THE BOUND COLLECTION
+            removeWarning(id);
+            // CALL A PROMISE THAT SHOULD BE IN A SERVICE
+            var promise = dismiss(id);
```
chrisvxd: Is this `promise` demo necessary? Seems we should just keep the demo simple

src/rb-warning-messages/demo/demo.tpl.html
```diff
@@ -0,0 +1,10 @@
+<div>
+    Some content might go here ...
+</div>
+<rb-warning-messages warnings="demoCtrl.warningList"
+    on-issue-clicked="demoCtrl.clicky(warning)"
```
chrisvxd: sorry to change this again... what do you think of `on-issue-click`, since we have `on-dismiss` below (not `on-dismissed`)

test/unit/rb-warning-messages/rb-warning-messages.spec.js
```diff
@@ -0,0 +1,99 @@
...
+
+                $scope.$apply();
+                warnings = angular.element(element.find('li'));
+                expect(warnings[0].outerHTML).toContain('ng-click="onIssueClicked(');
+                expect(warnings[1].outerHTML).toContain('ng-click="onDismiss(');
+            });
+
+            it('should attach an onDismiss function to the warnings', function () {
+                var list = angular.element(template),
+                    element = $compile(list)($scope),
```
chrisvxd: A lot of redefinition of variables happening here. Can we move these to the `beforeEach` method?

dev/app.js
nwhite89: think it will need to be as style_guide won't have this in it's app and therefore the demo won't work in there

src/rb-warning-messages/demo/demo-ctrl.js
darwinlopez: Demo is simple enough and easy to understand.... any doubt here ?

chrisvxd: It's not overly complex so let's leave it as it is

dev/app.js
chrisvxd: @darwinlopez ?

PR #54, opened 2015-04-21
Add `rb-page-title` demo docs

ohskylab:
TP: https://rockabox.tpondemand.com/entity/8849

Taken from style guide, revised to suit new usage.

Static version retained as reference for extending Angular version.

### DIFF
--- a/src/rb-page-title/demo/demo.tpl.html
+++ b/src/rb-page-title/demo/demo.tpl.html
@@ -1,3 +1,56 @@
-<rb-header>
-    <rb-page-title heading="My Test Page" subheading="This subheading rocks"></rb-page-title>
-</rb-header>
+<section class="Section Section--gutter">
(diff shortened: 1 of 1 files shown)


nwhite89:
you've removed the directive ?

ohskylab:
@nwhite89 [L21](https://github.com/rockabox/rbx_ui_components/pull/54/files#diff-7b68d9263c912fa49e3e5894133b737cR21) (but doesn't live in `<rb-header>` any more)

nwhite89:
ah klkl! :+1:

chrisvxd:
@ohskylab yeah but you've also removed the directive from the demo :P

chrisvxd:
@ohskylab wait okay so you have done but you've left the static demo in there too. Do we want to show both static and angular markup? Makes sense if so

PR #55, opened 2015-04-21
Add static overlay components

ohskylab:
* [x] `.OverlayModeless`
* [x] `.OverlayModal`


ohskylab:
> So the var `--stacking-system-message` is going to be applied to all `OverlayModeless` components? 

Yep. Unless we modify/update the stacking order. There's a case for changing the var name from `--stacking-system-message` to `--stacking-overlay-modeless`, I guess.

> Is the `--stacking-system-message` var being used elsewhere?

Not currently, and no plans to with what we can see.

chrisvxd:
@ohskylab scope creep to do as part of this?

ohskylab:
No, reasonable to change here. The `var()` def happens  in style guide and I have a couple of other changes for that, so will queue up some mini-PRs when I add the amend for this.

PR #58, opened 2015-04-22
Angularified rb-system-message

chrisvxd:
Also updated demo page to include directive API and angular examples.

Tests included. They use a new `compileTemplate` method instead of the `recompile` method to keep the tests more explicit. Here is a story in backlog for updating existing tests: https://rockabox.tpondemand.com/entity/8940.

TP https://rockabox.tpondemand.com/entity/8900

### DIFF
--- a/src/rb-system-message/demo/demo.tpl.html
+++ b/src/rb-system-message/demo/demo.tpl.html
@@ -2,95 +2,56 @@
     .SystemMessage
 </h1>
 
-<h2 class="u-textHeading2">
-    No modifier
-</h2>
+<h3>
+    API
+</h3>
 
-<div class="SystemMessage">
-    <div class="SystemMessage-inner">
-        <div class="SystemMessage-body">
-            <h2 class="u-textHeading2">
-                Message title
-            </h2>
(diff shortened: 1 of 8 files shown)


src/rb-system-message/demo/demo.tpl.html
```diff
@@ -2,95 +2,56 @@ <h1 class="u-textHeading1">
...
 
-<h2 class="u-textHeading2">
-    Positive modifier
-</h2>
+<rb-system-message
+    title="Something happened"
+    message="Here is the message"
+    technical-details="HUH"
+    dismiss-message="Custom onDismiss function"
+    on-dismiss="delete demoCtrl.messages[message.id]">
```
k0nG: This is causing an uncaught exception. Think you left it in here from our fiddling earlier.

chrisvxd: Oh shit yeah. Whoops.

chrisvxd: Oh yeah, whoops.

PR #63, opened 2015-04-22
Add Ratio of fraction with currency component

nwhite89:
- Allow fraction ratio's to be used with the rb-ratio component
- Allow the use of the rb-currency component when displaying fractions

TP: https://rockabox.tpondemand.com/entity/8805

### DIFF
--- a/src/rb-ratio/rb-ratio.tpl.html
+++ b/src/rb-ratio/rb-ratio.tpl.html
@@ -1,1 +1,9 @@
-<span class="Ratio">{{ ((numerator * 100) / denominator) || 0 | number:0 }}%</span>
+<span class="Ratio" ng-switch on="ratioType">
+    <span ng-switch-when="fraction" ng-switch on="displayType">
+        <span ng-switch-when="currency">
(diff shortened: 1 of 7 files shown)


src/rb-ratio/rb-ratio.tpl.html
```diff
@@ -1 +1,9 @@
-<span class="Ratio">{{ ((numerator * 100) / denominator) || 0 | number:0 }}%</span>
+<span class="Ratio" ng-switch on="ratioType">
+    <span ng-switch-when="fraction" ng-switch on="displayType">
+        <span ng-switch-when="currency">
+            <rb-currency amount="{{ numerator }}" decimal-places="{{ decimalPlaces }}"></rb-currency/>&nbsp;⁄ <rb-currency amount="{{ denominator }}" decimal-places="{{ decimalPlaces }}"></rb-currency>
+        </span>
+        <span ng-switch-default>{{ numerator }} ⁄ {{ denominator }}</span>
+    </span>
+    <span ng-switch-default>{{ ((numerator * 100) / denominator) || 0 | number:0 }}%</span>
+</span>
```
chrisvxd: Non-blocking: Could we add a couple of comments to describe the switching and make it easier to see what's going on?

PR #64, opened 2015-04-23
Ensure components that are dependent on others require them

nwhite89:
When a component needs to use another rb-component explicitly to render, ensure that the componenent requires via Webpack and via the Angular module dependency injection.


chrisvxd:
@nwhite89 merging but can we try and split out different tasks into different commits at the very least?

PR #67, opened 2015-04-23
FormMessage component

chrisvxd:
Added FormMessage component with `is-invalid` state.

Implemented in directive template already “angularified” (minus tests). Tests coming in next PR.

TP https://rockabox.tpondemand.com/entity/8955


src/rb-form-message/rb-form-message.css
```diff
@@ -0,0 +1,32 @@
...
+
+:root {
+    --FormMessage-border-radius: 4px;
+    --FormMessage-invalid-background-color: var(--color-red-base);
+    --FormMessage-invalid-color: var(--color-white);
+}
+
+/**
+ *
+ */
```
ohskylab: Ditch this empty comment block. Otherwise GTM.

PR #71, opened 2015-04-23
Creates-a-Progress-Button-attr-directive

darwinlopez:
Creates a progress button attribute directive that changes the default behaviour of a `<button>` from a single click to a click event that triggers an execution process that changes the button label between two
states (messages) every given time interval elapses.

Includes a extended demo to show how it could be used.

### DIFF
--- a/src/rb-progress-button/demo/demo.tpl.html
+++ b/src/rb-progress-button/demo/demo.tpl.html
@@ -1,64 +1,62 @@
-<div class="DataActionTable">
-    <table class="DataActionTable-main">
-        <thead>
-            <tr class="DataActionTable-row">
-                <th class="DataActionTable-cellHeader DataActionTable-colUtility DataActionTable-cellHeader--sortable u-textCenter" scope="col">
-                </th>
-                <th class="DataActionTable-cellHeader DataActionTable-cellHeader--sortable is-sortedDescending" scope="col">
-                    <a class="DataActionTable-cellHeaderItem">Status</a>
-                </th>
-                <th class="DataActionTable-cellHeader DataActionTable-cellHeader--sortable is-sortedDescending" scope="col">
-                    <a class="DataActionTable-cellHeaderItem">Row name</a>
-                </th>
-                <th class="DataActionTable-cellHeader DataActionTable-cellHeader--sortable is-sortedDescending" scope="col">
-                    <a class="DataActionTable-cellHeaderItem">Row description</a>
-                </th>
-                <th class="DataActionTable-cellHeader DataActionTable-colUtility">
-                    <span class="DataActionTable-cellHeaderItem u-hiddenVisually">
-                        Line actions
-                    </span>
-                </th>
-            </tr>
-        </thead>
-        <tbody class="DataActionTable-body">
-            <tr class="DataActionTable-row" ng-repeat="item in demoCtrl.itemList track by $index">
-                <td class="DataActionTable-cellData u-textCenter">
-                    <div class="DataActionTable-cellDataItem">
-                        <rb-badge state="status{{item.status}}" collapsed="true"></rb-badge>
-                    </div>
-                </td>
-                <th class="DataActionTable-cellData" scope="row">
-                    <div class="DataActionTable-cellDataItem">
-                        {{item.status}}
-                    </div>
-                </th>
-                <th class="DataActionTable-cellData" scope="row">
-                    <div class="DataActionTable-cellDataItem">
-                        <a href="#">{{item.name}}</a>
-                    </div>
-                </th>
-                <th class="DataActionTable-cellData" scope="row">
-                    <div class="DataActionTable-cellDataItem">
-                        <a href="#">{{item.description}}</a>
-                    </div>
-                </th>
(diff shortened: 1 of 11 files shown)


src/rb-progress-button/demo/demo.tpl.html
```diff
@@ -0,0 +1,64 @@
...
+                <th class="DataActionTable-cellData" scope="row">
+                    <div class="DataActionTable-cellDataItem">
+                        <a href="#">{{item.name}}</a>
+                    </div>
+                </th>
+                <th class="DataActionTable-cellData" scope="row">
+                    <div class="DataActionTable-cellDataItem">
+                        <a href="#">{{item.description}}</a>
+                    </div>
+                </th>
```
chrisvxd: Can we strip out `.DataActionTable` from the demo and just show the buttons?

test/unit/rb-progress-button/rb-progress-button.tpl.html
```diff
@@ -0,0 +1,4 @@
+<button rb-progress-button on-execute="executeFunction()"
```
chrisvxd: Can you use this on `rb-button`?

chrisvxd:
@darwinlopez too late to finish reviewing this now so will continue on monday unless @k0nG feels the desire to do it tomorrow!

src/rb-progress-button/rb-progress-button-directive.js
```diff
@@ -0,0 +1,150 @@
...
+                // Default action message
+                actionLabel = 'Executing',
+                // Default progress message
+                progressLabel = 'Please Wait';
+
+            if (angular.isDefined(attr.interval) &&
+                !isNaN(parseInt(attr.interval)) &&
+                isFinite(attr.interval)) {
+
+                interval = attr.interval;
```
chrisvxd: Can you indent this?

src/rb-progress-button/rb-progress-button-directive.js
```diff
@@ -0,0 +1,150 @@
...
+                originalLabel,
+                currentLabel,
+                // Default interval of 2000 milliseconds
+                interval = 2000,
+                // Default action message
+                actionLabel = 'Executing',
+                // Default progress message
+                progressLabel = 'Please Wait';
+
+            if (angular.isDefined(attr.interval) &&
```
chrisvxd: Can you add a line comment to describe what this if statement does?

src/rb-progress-button/rb-progress-button-directive.js
```diff
@@ -0,0 +1,150 @@
...
+                elem[0].blur();
+
+                stopInterval = $interval(function () {
+                    if (currentLabel === progressLabel) {
+                        currentLabel = actionLabel;
+                    } else {
+                        currentLabel = progressLabel;
+                    }
+                    elem.text(currentLabel);
+                }, interval);
```
chrisvxd: Why is this called `stopInterval`?

src/rb-progress-button/rb-progress-button-directive.js
```diff
@@ -0,0 +1,150 @@
...
+
+            elem.on('click', function (event) {
+
+                // Execution not allowed if any execution is under way
+                // or execute function is not present
+                if (angular.isDefined(stopInterval) ||
+                    angular.isUndefined(attr.onExecute) ||
+                    angular.equals(attr.onExecute, '')) {
+
+                    return;
```
chrisvxd: Indent, please :)

src/rb-progress-button/demo/demo.tpl.html
darwinlopez: We shouldn't downgrade the demo as it provides with some info and context on how to integrate the component.

test/unit/rb-progress-button/rb-progress-button.tpl.html
darwinlopez: Is this a restriction on the attr. directive ?

chrisvxd: Yes since the `progress-button` is an extension of `rb-button`. For example, the styling is dependant on the styles in `rb-button`. The fact that it works on any button is just coincidence IMO!

src/rb-progress-button/demo/demo.tpl.html
chrisvxd: Context is great but there is a load of code in here that doesn't do anything- `DataActionTable` doesn't exist in this repo. Do you think the extra context is necessary? My feeling is that the progress button's behaviour is self explanatory to what it does.

src/rb-progress-button/rb-progress-button-directive.js
darwinlopez: A common convention for intervals and timeouts in angular

chrisvxd: Do you think it makes sense?

darwinlopez: Yup, and something similar happens when cleaning watchers as well

PR #73, opened 2015-04-24
Remove spinner on disabled `.TextControl` number inputs

ohskylab:
Conflicts with the disabled padlock icon and not needed when control is disabled.


ohskylab:
This is prep for the `.TextControl--currency` modifier and it's exciting multiple background nature.

jaylett:
LGTM: Squash then I'll merge.

PR #75, opened 2015-04-24
Add `.TextControl--currency` modifier

ohskylab:
TP: https://rockabox.tpondemand.com/entity/8654

Unlike `.TextControl--search`, this field requires validation so is more complex.

* Uses CSS multiple backgrounds.
* Creates variables for common backgrounds and position patterns used in no-modifier version.
* Creates new `.is-populated` state that should be set on fields by Angular, as this can't be done via CSS.

Can be used as a template for similar types in future.


src/rb-text-control/demo/demo.tpl.html
```diff
@@ -1,8 +1,39 @@
...
+            <p>
+                Matches <code>:focus</code> state. There's no way of changing an <code>input</code> style if it's
+                populated (or not). This adds the correct background image and needs setting via Angular.
+            </p>
+
+            <div class="TextControl TextControl--currency is-populated">
+                <label class="TextControl-label is-populated u-textLabel" for="f-Text-currency-populated">Currency label</label>
+                <input class="TextControl-field is-populated" id="f-Text-currency-populated" name="number" placeholder="Currency" type="number" value="666" required>
+            </div>
+
```
chrisvxd: Not super happy to merge this without some clearer separation in the demo between static and angular text controls- the demo now looks super messy.

chrisvxd:
What a sexy control. Just needs demo sorting out @ohskylab

ohskylab:
@chrisvxd Updated demo to match the structure I've used for `rb-page-title` i.e. breaking Angular and Static parts out into separate sections, Angular first. OK?

PR #82, opened 2015-04-28
Login form message.

k0nG:
Message passed on login form attribute used to show rb-form-message.

### DIFF
--- a/src/rb-login-form/demo/demo.tpl.html
+++ b/src/rb-login-form/demo/demo.tpl.html
@@ -5,7 +5,7 @@
             email-model="demoCtrl.email"
             password-model="demoCtrl.password"
             on-login="demoCtrl.login()"
-            message="demoCtrl.message">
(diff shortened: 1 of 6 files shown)


src/rb-login-form/demo/demo.tpl.html
```diff
@@ -4,7 +4,8 @@
             icon="white-alabaster-175-logo-rockabox"
             email-model="demoCtrl.email"
             password-model="demoCtrl.password"
-            on-login="demoCtrl.login()">
+            on-login="demoCtrl.login()"
+            message="demoCtrl.message">
```
chrisvxd: I don't think this should be an expression.... probably an evaluated one, i.e. `{{ demoCtrl.message }}`

k0nG: Yeah that makes sense. Then we can use it without a controller.

PR #86, opened 2015-04-29
Add Autoprefixer `browserslist` config for PostCSS

ohskylab:
Same version as in rockabox/rbx_style_guide.

See: https://github.com/ai/browserslist


ohskylab:
Matching Campaign Management PR at rockabox/rbx_campaign_management/pull/51.

PR #93, opened 2015-04-29
Adds the possibility of using rb-button in the page title component

nwhite89:
TP: https://rockabox.tpondemand.com/entity/9106


ohskylab:
Does the Angular version now do everything that the static version does? If so, should probably update the demo (drop the static version, remove the Angular heading).

nwhite89:
Yeah me and @chrisvxd discussed this thought keep it so you can change when additions needed? Up to you

ohskylab:
Ah, appreciate it, but I reckon remove it and we can fish a static out of the history if/when we need it.

PR #94, opened 2015-04-30
`.RadioControl` WIP

ohskylab:
TP: https://rockabox.tpondemand.com/entity/9093

See other half of this pair at #96.

Read the docs in each demo for details, but note the following in particular:
 
* No styles yet besides a row modifier for radio buttons.
* RadioControl must logically always be a set of of field/label pairings.
* CheckControl should structurally be a single field/label pair, for re-use.
* Help and Invalid messages should be surfaced on the group for RadioControl, but on the parent (fieldset?) for CheckControl.
* Will need new components groupings (e.g. `rb-multi-check` for a group of checkboxes topped by a "select all children" checkbox).
* Plan is to use CSS multicol for columns of checkboxes, for automatic flowing into columns and a sane source order. See: http://codepen.io/anon/pen/jPPVye


ohskylab:
Mergeable if everyone agrees that this is the right way to go. If it goes in, please create a new story to for full styling.

PR #96, opened 2015-04-30
`.CheckControl` WIP

ohskylab:
TP: https://rockabox.tpondemand.com/entity/9095

See other half of this pair at #94, including PR notes for both.


ohskylab:
Mergeable if everyone agrees that this is the right way to go. If it goes in, please create a new story to for full styling.

PR #98, opened 2015-04-30
Add angular-elastic to the textarea

nwhite89:
Use Angular-elastic and make it so that it is *not* required

TP: https://rockabox.tpondemand.com/entity/9086

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -31,5 +31,8 @@
   },
   "resolutions": {
     "angular": "~1.3.15"
+  },
+  "dependencies": {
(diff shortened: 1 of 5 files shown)


bower.json
```diff
@@ -31,5 +31,8 @@
   },
   "resolutions": {
     "angular": "~1.3.15"
+  },
+  "dependencies": {
+    "angular-elastic": "rockabox/angular-elastic#master"
```
farridav: why are you not using a tag?

nwhite89: as discussed, needed until patch of ours is in angular-elastic

PR #99, opened 2015-04-30
Task/9034 modal overlay

darwinlopez:
Creates a modal overlay that wraps a modal content and prevents interaction behind the modal window.
This task completes the story 8913.


src/rb-overlay-modal/rb-overlay-modal-directive.js
```diff
@@ -22,12 +25,44 @@ define([
...
-            },
             restrict: 'E',
-            replace: true,
-            template: template
+            transclude: true,
+            template: template,
+            link: link
         };
+
+        function link (scope, elem, attr) {
```
nwhite89: can we split the link function out into it's own and load it in via webpack please

PR #100, opened 2015-04-30
`.Fieldset` WIP

ohskylab:
TP: https://rockabox.tpondemand.com/entity/8113

This is DNM because it's not totally finished and tested (see TP), but OK for it to go through if it's needed over the next couple of weeks. If you merge it, please create follow-up stories (static and Angular) to fix it properly.


nwhite89:
Merged due to blocking TP story still open looks like mainly CSS work rather than Angular work needed @ohskylab

PR #101, opened 2015-04-30
rb section as angular directive.

k0nG:
TP: https://rockabox.tpondemand.com/entity/9071

RB section directive with modifiers.

### DIFF
--- a/src/rb-section/rb-section-directive.js
+++ b/src/rb-section/rb-section-directive.js
@@ -23,9 +23,13 @@ define([
 
         return {
             scope: {
+                title: '@',
+                gutterless: '=',
+                hideTitle: '='
             },
             restrict: 'E',
             replace: true,
+            transclude: true,
             template: template
         };
     }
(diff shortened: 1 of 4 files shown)


src/rb-section/rb-section-directive.js
```diff
@@ -23,9 +23,13 @@ define([
 
         return {
             scope: {
+                title: '@',
+                gutterless: '=',
```
nwhite89: should be ``'@'`` as we don't need to amend the variable within the directive?

src/rb-section/rb-section-directive.js
```diff
@@ -23,9 +23,13 @@ define([
...
+                hideTitle: '='
```
nwhite89: same here this should be ``'@'`` as we don't need to amend the variable within the directive?

src/rb-section/rb-section.tpl.html
```diff
@@ -0,0 +1,4 @@
+<section class="Section" ng-class="{'Section--gutterless': gutterless}">
+    <h2 class="u-textHeading2" ng-class="{'u-hiddenVisually': hideTitle}">{{ title }}</h2>
```
nwhite89: ``{{::title}}``

PR #102, opened 2015-05-01
Modal confirm directive.

k0nG:
TP: https://rockabox.tpondemand.com/entity/8811

- Transcludes contents for message.
- Title as attribute.
- `on-cancel` and `on-confirm` attributes pass callbacks to buttons.


src/rb-modal-confirm/rb-modal-confirm.tpl.html
```diff
@@ -0,0 +1,20 @@
+<div class="ModalConfirm">
+    <div class="ModalConfirm-title">
+        <h2 class="u-textHeading2">{{ title }}</h2>
```
nwhite89: could you make this ``{{::title}}`` please

PR #103, opened 2015-05-01
rb-text-control currency field

nwhite89:
TP: https://rockabox.tpondemand.com/entity/9083

### DIFF
--- a/test/unit/rb-text-control/rb-text-control-currency.spec.js
+++ b/test/unit/rb-text-control/rb-text-control-currency.spec.js
@@ -1,0 +1,84 @@
+define([
+    'components/rb-text-control',
+    'html!./rb-text-control-currency.tpl.html'
+], function (rbTextControl, template) {
+    describe('rb-text-control', function () {
+
+        var $rootScope,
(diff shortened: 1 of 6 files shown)


test/unit/rb-text-control/rb-text-control-currency.spec.js
```diff
@@ -0,0 +1,84 @@
...
+                    element.find('input').triggerHandler('blur');
+
+                    // Get validation message
+                    element = angular.element(element[0].querySelector('.TextControl-message.is-invalid'));
+
+                    expect(element.html()).toContain('Please enter a valid monetary value.');
+                });
+            });
+
+            describe('accepts', function () {
```
k0nG: Think this test should also test `102.13` as that's a valid value with decimals. From these tests you can only assert that that a valid value is `102` and an invalid `102.122`, which leaves ambiguity around a value with two decimal places.

nwhite89: ah yes good point will do

PR #104, opened 2015-05-01
Angularify the rb-fieldset component

nwhite89:
Moves the Fieldset component to be "Angularified"

TP: https://rockabox.tpondemand.com/entity/9081

### DIFF
--- a/test/unit/rb-fieldset/rb-fieldset.spec.js
+++ b/test/unit/rb-fieldset/rb-fieldset.spec.js
@@ -25,11 +25,48 @@ define([
             };
         }));
 
-        describe('', function () {
-            it('should ', function () {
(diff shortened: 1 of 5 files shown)


test/unit/rb-fieldset/rb-fieldset.spec.js
```diff
@@ -25,11 +25,45 @@ define([
             };
         }));
 
-        describe('', function () {
-            it('should ', function () {
-                compileTemplate('<rb-fieldset></rb-fieldset>');
-            });
+        it('should add a title to the component', function () {
+            compileTemplate('<rb-fieldset title="Default Text"></rb-fieldset>');
```
k0nG: I'd check that that text is put into the `<div class="Fieldset-title"><span>` .

PR #108, opened 2015-05-05
Include overlay modal dependancy in modal confirm

k0nG:
Need to include overlay modal as a dependancy.


src/rb-modal-confirm/index.js
```diff
@@ -1,7 +1,8 @@
 define([
+    '../rb-overlay-modal',
```
nwhite89: can you change this to use the webpack aliases please ``components/rb-overlay-modal``

PR #112, opened 2015-05-06
rb-select component

nwhite89:
- Add's select to use ``ng-required`` (allowing to set it as a required field)
- Add's title and help text to select
- Add's ``is-disabled`` to the ``rb-select``

TP: https://rockabox.tpondemand.com/entity/9115


k0nG:
Can you update your tests to cover all these attributes. See https://github.com/rockabox/rbx_ui_components/blob/master/test/unit/rb-modal-confirm/rb-modal-confirm.spec.js#L115 for examples.

PR #113, opened 2015-05-06
Datetime control component

k0nG:
TP: https://rockabox.tpondemand.com/entity/9085

- Using same date time directive from rbx_ui: http://mgcrea.github.io/angular-strap/#/datepickers
- Pulling custom bootstrap css generated by http://getbootstrap.com/customize/ to just include things we need to datepicker.
- Including extra datepicker css from https://github.com/mgcrea/bootstrap-additions/blob/master/src/datepicker/datepicker.less
- Added `eot` file loading to webpack.
- Changed `grunt test` task to export `TZ=UTC` environmental var so phantomJs runs in UTC timezone and therefore produces predictable results when testing time inputs.
- Added help text on input to tell user that time presented in in their local browser time.


src/rb-datetime-control/rb-datetime-control.css
```diff
@@ -0,0 +1,1709 @@
...
+  font-weight: bold;
+}
+table {
+  border-collapse: collapse;
+  border-spacing: 0;
+}
+td,
+th {
+  padding: 0;
+}
```
nwhite89: I think everything between this line and Line 18 should be removed as it's going to override our own normalise stuff added by @ohskylab

src/rb-datetime-control/rb-datetime-control.tpl.html
```diff
@@ -0,0 +1,51 @@
...
+        ng-disabled="{{isDisabled}}"
+        ng-readonly="{{isReadonly}}"
+        ng-class="{'is-invalid': form['date'].$touched && form['date'].$invalid}"
+        placeholder="{{placeholderDate}}"
+        autocomplete="off"
+        bs-datepicker>
+    <input
+        type="text"
+        class="TextControl-field"
+        name="time"
```
nwhite89: Doesn't this make it so you can't have multiple date time's for the same form?

src/rb-datetime-control/demo/demo.tpl.html
```diff
@@ -0,0 +1,54 @@
...
+        <h2>
+            rb-datetime-control
+        </h2>
+        <h3>
+            Datetime Control Angular Version
+        </h3>
+    </div>
+    <div class="ComponentView">
+        <h4>Basic usage</h4>
+        <form name="basicForm">
```
nwhite89: Can you use one form for the demo, to ensure that you can have multiple date time field's on one form please

Gruntfile.js
```diff
@@ -12,12 +12,18 @@ module.exports = function (grunt) {
...
         var tasks = [
             'lint',
+            'env:testUTC',
             'karma:unit'
         ];
 
         grunt.task.run(tasks);
     });
 
+    grunt.registerTask('unit', [
```
nwhite89: can you update the README please

src/rb-datetime-control/rb-datetime-control.css
k0nG: Yeah agreed. Will remove as much as I can.

Gruntfile.js
k0nG: Done.

dev/app.js
```diff
@@ -36,6 +37,7 @@ define([
     rbButtonDemo,
     rbCurrencyDemo,
     rbDataSummaryDemo,
+    rbDatetimeControl,
```
chrisvxd: @k0nG this is the `Demo`

nwhite89:
kl I'm happy for merge would like @chrisvxd to sign off also as added more comment's than me

PR #115, opened 2015-05-07
Allow for setting target

nwhite89:
When setting the href of a nav bar item allow it to be possible to set the target to something like ``_self`` for when needing the ui-router to no longer listen to the change as a state.

TP: https://rockabox.tpondemand.com/entity/9073


k0nG:
I might be wrong and this work might be needed. But when I worked on this before just setting an `href` and not a `ui-sref` was enough for the link to be ignored by `ui-router` I'm not sure `_self` is required?

k0nG:
@nwhite89  explained this is needed.

PR #116, opened 2015-05-07
TextControl-field in rb-select so there is validation message padding

chrisvxd:


src/rb-select/rb-select.tpl.html
```diff
@@ -5,6 +5,7 @@
         {{ label }}
     </label>
     <select
+        class="TextControl-field"
```
nwhite89: I don't think we should use this, I added and it messed with the styling I'd rather not and let @ohskylab decide after

PR #123, opened 2015-05-07
Angular radio component

jonman:
TP: https://rockabox.tpondemand.com/entity/9094

### DIFF
--- a/src/rb-radio-control/rb-radio-control.tpl.html
+++ b/src/rb-radio-control/rb-radio-control.tpl.html
@@ -1,13 +1,13 @@
 <div class="RadioControl" ng-class="{
-    'RadioControl--row': row
+    'RadioControl--row': isRow
 }">
 
     <div class="RadioControl-items">
         <div
             class="RadioControl-item"
             ng-class="{
-                'is-invalid': form[name].$touched && form[name].$invalid,
                 'is-disabled': isDisabled,
+                'is-invalid': form[name].$touched && form[name].$invalid,
                 'is-required': isRequired
             }"
             ng-repeat="(key, value) in data">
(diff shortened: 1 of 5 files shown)


src/rb-radio-control/rb-radio-control.tpl.html
```diff
@@ -0,0 +1,49 @@
...
+            }"
+            ng-repeat="(key, value) in data">
+
+            <input
+                class="RadioControl-field"
+                id="{{name}}-{{key}}"
+                name="{{name}}"
+                ng-checked="{{value.checked}}"
+                ng-class="{
+                    'is-focused': value.focused,
```
nwhite89: I'd rather that we didn't do this, it adds extra watchers to the component via Angular as it will watch for the value to change, I don't think we should do something like that purely for a demo purpose?

src/rb-radio-control/rb-radio-control.tpl.html
```diff
@@ -0,0 +1,49 @@
...
+                class="RadioControl-label u-textLabel"
+                for="{{name}}-{{key}}">
+                {{value.label}}
+            </label>
+        </div>
+    </div>
+
+    <div
+        class="RadioControl-message"
+        ng-show="helpMessage">
```
nwhite89: can you change this to ng-if='" please, means that the element won't be added to the DOM at all

src/rb-radio-control/rb-radio-control.tpl.html
```diff
@@ -0,0 +1,49 @@
...
+                'is-disabled': isDisabled,
+                'is-invalid': form[name].$touched && form[name].$invalid,
+                'is-required': isRequired
+            }"
+            ng-repeat="(key, value) in data">
+
+            <input
+                class="RadioControl-field"
+                id="{{name}}-{{key}}"
+                name="{{name}}"
```
nwhite89: we may not be doing this else where @chrisvxd may have thoughts on this so not blocking PR perhaps a thought for a small refactor over all of these form elements but I think name for example should be displayed with ``{{::name}}`` preventing Angular from watching the value, not sure we will need to ever change name's ?

src/rb-radio-control/rb-radio-control.tpl.html
jonman: I'm not if it is only for demo purposes though, so for now I'll remove them and will ask @ohskylab when he gets back.

nwhite89: cool I'd imagine it is? but yeah let @ohskylab have sign-off cheers

PR #124, opened 2015-05-08
Have rb-select return a single value not obj

nwhite89:
TP: https://rockabox.tpondemand.com/entity/9089


k0nG:
Can you also update your demo's `previouslySelected` variable to set a key not the whole object. Just for sake of consistency.

PR #128, opened 2015-05-14
Date time inherit from another source.

k0nG:
TP: https://rockabox.tpondemand.com/entity/9321

Sets date time from a given source when a checkbox is checked.


chrisvxd:
The checkbox also needs to disable the input fields. This is both from a UX point of view, but also technically we don't want to be modifying the inherited model.

PR #130, opened 2015-05-15
Angular checkbox component

jonman:
TP: https://rockabox.tpondemand.com/entity/9096

### DIFF
--- a/src/rb-check-control/demo/demo-ctrl.js
+++ b/src/rb-check-control/demo/demo-ctrl.js
@@ -3,13 +3,13 @@ define([
 
     // @ngInject
     function demoCtrl ($rootScope, $state, $injector) {
-        this.data = {
-            'default': [{label: 'Default label'}],
-            'invalid': [{label: 'Invalid label'}],
-            'disabled': [{label: 'Disabled label', disabled: true}],
-            'required': [{label: 'Required label', required: true}],
-            'checked': [{label: 'Checked label', checked: true}]
-        };
+        this.ngModel = [
+            {label: 'Default label'},
+            {label: 'Invalid label'},
+            {label: 'Disabled label', disabled: true},
+            {label: 'Required label', required: true},
+            {label: 'Checked label', checked: true}
+        ];
     }
 
     return demoCtrl;
--- a/src/rb-check-control/demo/demo-ctrl.js
+++ b/src/rb-check-control/demo/demo-ctrl.js
@@ -3,10 +3,16 @@ define([
 
     // @ngInject
     function demoCtrl ($rootScope, $state, $injector) {
-        this.ngModel = [
+        this.default = [
             {label: 'Default label'},
             {label: 'Invalid label'},
             {label: 'Disabled label', disabled: true},
+            {label: 'Required label', required: true},
+            {label: 'Checked label', checked: true}
+        ];
+        this.selectAll = [
+            {label: 'Default label'},
+            {label: 'Invalid label'},
             {label: 'Required label', required: true},
             {label: 'Checked label', checked: true}
         ];
--- a/src/rb-check-control/demo/demo-ctrl.js
+++ b/src/rb-check-control/demo/demo-ctrl.js
@@ -3,6 +3,13 @@ define([
 
     // @ngInject
     function demoCtrl ($rootScope, $state, $injector) {
+        this.data = {
+            'default': [{label: 'Default label'}],
+            'invalid': [{label: 'Invalid label'}],
+            'disabled': [{label: 'Disabled label', disabled: true}],
+            'required': [{label: 'Required label', required: true}],
+            'checked': [{label: 'Checked label', checked: true}]
+        };
     }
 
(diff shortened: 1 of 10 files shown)


src/rb-check-control/rb-check-control-select-all-directive.js
```diff
@@ -0,0 +1,87 @@
+define([
+    'html!./rb-check-control-select-all.tpl.html'
+], function (template) {
+
+    /**
+     * @ngdoc directive
+     * @name rbCheckControlSelectAll
+     * @module rb-check-control0-
```
k0nG: whats this 0 from?

src/rb-check-control/rb-check-control-select-all-directive.js
```diff
@@ -0,0 +1,87 @@
...
+     *
+     * @restrict E
+     *
+     * @description
+     *
+     * @usage
+     * <hljs lang="html">
+     *    <rb-check-control>
```
k0nG: usage example.

src/rb-check-control/rb-check-control-select-all-directive.js
jonman: Typo!

src/rb-check-control/rb-check-control-select-all-directive.js
```diff
@@ -0,0 +1,94 @@
...
+                    angular.forEach($scope.ngModel, function (value, key) {
+                        if (value.checked) {
+                            cleared = false;
+                        } else {
+                            checked = false;
+                        }
+                    });
+
+                    if (checked) {
+                        $scope.master = true;
```
chrisvxd: Not a huge fan of calling this `master`

src/rb-check-control/rb-check-control-select-all.tpl.html
```diff
@@ -0,0 +1,12 @@
+<div>
+    <input
+        id="{{::name}}-title"
+        ng-change="change()"
+        ng-disabled="{{::isDisabled}}"
+        ng-show="{{::selectAll}}"
```
chrisvxd: What do you say to putting this in the `rb-check-control.tpl.html` template? Why should we render the directive if we don't want to select all? Especially since this is a one-way binding.

src/rb-check-control/rb-check-control.tpl.html
```diff
@@ -0,0 +1,39 @@
+<div>
+    <rb-check-control-select-all
+        name="{{::name}}"
+        is-disabled="{{::isDisabled}}"
+        is-required="{{::isRequired}}"
+        ng-model="ngModel"
+        select-all="{{::selectAll}}"
```
chrisvxd: I would prefer this attr to have a more explicit name... currently writing up some guidelines for directive attrs

src/rb-check-control/demo/demo-ctrl.js
```diff
@@ -3,6 +3,19 @@ define([
...
+            {label: 'Disabled label', disabled: true},
+            {label: 'Required label', required: true},
+            {label: 'Checked label', checked: true}
+        ];
+        this.selectAll = [
+            {label: 'Default label'},
+            {label: 'Invalid label'},
+            {label: 'Required label', required: true},
+            {label: 'Checked label', checked: true}
+        ];
```
chrisvxd: How well is this model structure going to fit in with our API?

src/rb-check-control/rb-check-control-select-all-directive.js
jonman: Any suggestion? `title`, `parent`, `toggleSelectAll` etc ..

src/rb-check-control/rb-check-control-select-all.tpl.html
jonman: I originally had it within the same template but things got messy when differentiating between the title checkbox and child checkboxes during unit tests and element lookups within the directive. It seemed neater at the time to separate the two.

`$element.find('input')` would give me both the select all and children. If I wanted to do anything with the select all we would have to wait for the document to load then do a normal `document.getElementById`.

src/rb-check-control/rb-check-control.tpl.html
jonman: Sure, makes sense. Will change to `enable-select-all`

src/rb-check-control/rb-check-control-select-all-directive.js
jonman: `isSelected` agreed

src/rb-check-control/demo/demo-ctrl.js
chrisvxd: Discussed- will be fine like this

src/rb-check-control/rb-check-control-select-all.tpl.html
chrisvxd: Agreed to leave separately

PR #139, opened 2015-05-20
Clone `rb-select` to `rb-select-control`

ohskylab:
TP: https://rockabox.tpondemand.com/entity/9373

Be good for someone with strong ng chops than me to check I've remembered everything.

Cloning to a new component rather than renaming the existing component means we can hold off doing a major release of ui_components for a little while.


src/rb-select-control/demo/demo.tpl.html
```diff
@@ -0,0 +1,82 @@
...
+        </div>
+        <div class="ComponentView">
+            <rb-select-control
+                form="userForm"
+                label="Select field no previous (with a placeholder)"
+                value="id"
+                display="name"
+                name="noPreviousPlaceholder"
+                is-required="true"
+                items="demoRbSelectCtrl.items"
```
chrisvxd: The demo is broken. You need to find/replace `demoRbSelectCtrl` with `demoRbSelectControlCtrl` (see above)

src/rb-select-control/demo/demo-state.js
```diff
@@ -0,0 +1,15 @@
+define([
+    'html!./demo.tpl.html'
+], function (template) {
+
+    // @ngInject
+    function demoState ($stateProvider, $urlRouterProvider) {
+        $stateProvider.state('rb-select-control', {
+            url: '/rb-select-control',
+            controller: 'demo-rb-select-control-ctrl as demoCtrl',
```
chrisvxd: Can you swap `demoCtrl` with `demoRbSelectControlCtrl`? (see below)

ohskylab:
Like so? If it's good, I'll squash before merge (more conscious of this now I have to do release notes!)

PR #140, opened 2015-05-20
Creates a sidebar with scrollspy

darwinlopez:
Creates a sidebar with scrollspy

Because the angular-strap scrollspy is not able to keep track of changes in the target list, it was necessary to find a "good-enough" solution for this.
The outcome from a conversation with Anna and Rob was to remove (from the DOM) the scrollspy every time a user inputs a char in the search field and append it (back to the DOM) whenever the search field becomes empty.

TP: https://rockabox.tpondemand.com/entity/9298


k0nG:
:+1: LGTM

dev/app.js
```diff
@@ -29,6 +29,7 @@ define([
     'components/rb-section/demo',
     'components/rb-select/demo',
     'components/rb-side-nav/demo',
+    'components/rb-sidebar-scrollspy/demo',
```
chrisvxd: The name of the component should either be `sidebar` or `scrollspy`... speak to @ohskylab!

Need to get this right otherwise it becomes a breaking change

ohskylab: My vote: `scrollspy`. It shouldn't be explicitly tied to sidebar.

chrisvxd:
@darwinlopez part way through reviewing this, will continue tomorrow. Please do not merge yet.

dev/app.js
darwinlopez: Guys: bear in mind this is not a sidebar neither a scrollspy.... this is a sidebar using a scrollspy.
- The scrollspy is provided by the directive "bs-scrollspy" shipped in angular-strap... 
- This is tied to a sidebar because is using a template for a sidebar that makes use of the scrollspy
- In case you want to use the scrollspy with a navBar or tabs, you should build another directive with a navBar or Tab template using the "bs-scrollspy" or have this directive to make dynamic templating.

chrisvxd:
Tests need fixing but reprioritising towards sprint goal for now. Will pickup on Monday unless someone else wants to fix tests tomorrow

src/rb-scrollspy/rb-scrollspy-directive.js
```diff
@@ -0,0 +1,50 @@
...
+     *          {
+     *              label: 'Text to be displayed in the sidebar'
+     *              anchor: 'HTML anchor name. This anchor represents the target
+     *                       selector and is expected to match an id in another list.
+     *                       If no present or empty the item will not target any id.'
+     *              category: 'If `true` the item label should be displayed and
+     *                        styled as a category'.
+     *          }
+     *  'offset': Pixels to offset from top of screen when calculating position of
+     *            scroll. Defaults to '0'.
```
chrisvxd: Note to self... this documentation is now out of date

ohskylab:
Please update changelog on this before merge.

chrisvxd:
Can we make sure test commits are squashed and the commit name before merge please?

PR #145, opened 2015-05-22
Update README

ohskylab:
The current version is inaccurate, unhelpful and just plain wrong in places. And it's quite public.

Before merge, this would benefit from:

* [x] Core UI team member to check accuracy (@nwhite89, @k0nG, @chrisvxd)
* [x] [Language check](https://github.com/rockabox/rbx_ui_components/blob/paul/update-readme/README.md) (@jaylett, if you have time?)
* [x] Non-UI team member who doesn't already have ui_components installed to quickly run through [revised instructions](https://github.com/rockabox/rbx_ui_components/blob/paul/update-readme/README.md) to check they actually work (@mimpse, @jamescooke, @JoseKilo, @robertweglarek, @facundfortuny - any volunteers?)
* [x] Serious squashing (@ohskylab)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,53 +1,57 @@
-# RBX UI Components
+# RBX UI Components [![Build Status](https://travis-ci.org/rockabox/rbx_ui_components.svg?branch=master)](https://travis-ci.org/rockabox/rbx_ui_components)
 
-A set of angular components.
+Rockabox's [AngularJS](https://angularjs.org/) and [PostCSS](https://github.com/postcss/postcss) UI components.
 
-## Dependencies
+## Getting Started
 
-* [Git](http://git-scm.com)
-* [Grunt](http://gruntjs.com/): `npm install -g grunt`.
-* [Bower](http://bower.io): `npm install -g bower`.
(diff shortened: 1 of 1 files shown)


k0nG:
:+1: Nice tidy up.

mimpse:
Works for me. One trifling thing that's a result of my having worked in publishing, and which I can't comment on in context as Github's inline commenting seems broken: line 38,

`Unit tests should not be run directly with Karma`

Would benefit from the active voice - on first reading, it implied that unit tests running in Karma was unlikely, so perhaps:

`You should not run unit tests directly...`

Picky, I know...

jaylett:
I'd go further than @mimpse, and say something like:

```
Do not run unit tests directly with Karma; the grunt tasks are required to set the timezone correctly for PhantomJS.
```

nwhite89:
@jaylett @mimpse makes sense @ohskylab want to address then merge? :+1:

ohskylab:
@jaylett @mimpse Agree with that, it was terribly passive. I've reworded and squashed all commits.

Good to MOG afaic!

PR #146, opened 2015-05-22
Add `.SelectControl` styles

ohskylab:
TP: https://rockabox.tpondemand.com/entity/9179

Temporary, to tide us over until a custom typeahead is built.

Uses the exciting, new(ish) `appearance` property. This should work in Chrome, Firefox,
other Webkit but may fail horribly in IE. C'est la vie.

### DIFF
--- a/src/rb-select-control/demo/demo.tpl.html
+++ b/src/rb-select-control/demo/demo.tpl.html
@@ -1,10 +1,94 @@
-
 <section class="Section">
     <form name="userForm">
         <div class="RawHTML">
             <h2>
                 rb-select-control
             </h2>
+            <h3>
+                Static Test
+            </h3>
+        </div>
+        <div class="ComponentView">
+            <div class="SelectControl">
+                <label class="SelectControl-label u-textLabel" for="SelectDemo">
+                    Normal
+                </label>
+                <select class="SelectControl-field" id="SelectDemo">
+                    <option value="">Car</option>
+                    <option value="0" label="Volvo">Volvo</option>
+                    <option value="1" label="Audi">Audi</option>
+                    <option value="2" label="BMW">BMW</option>
+                    <option value="3" label="Mercedes">Mercedes</option>
+                </select>
+                <div class="SelectControl-message">
+                    i.e. Your favourite make
+                </div>
+            </div>
+            <div class="SelectControl SelectControl--small">
+                <label class="SelectControl-label u-textLabel" for="SelectDemo">
+                    Small
+                </label>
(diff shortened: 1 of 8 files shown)


src/rb-select-control/demo/demo.tpl.html
```diff
@@ -1,10 +1,94 @@
...
+                </label>
+                <select class="SelectControl-field is-invalid" id="SelectDemoInvalid">
+                    <option value="">Car</option>
+                    <option value="0" label="Volvo">Volvo</option>
+                    <option value="1" label="Audi">Audi</option>
+                    <option value="2" label="BMW">BMW</option>
+                    <option value="3" label="Mercedes">Mercedes</option>
+                </select>
+                <div class="SelectControl-message is-invalid">
+                    i.e. Your favourite make
```
chrisvxd: This is a weird example `is-invalid` message

chrisvxd:
Can't help but feel like the dropdown arrow should react when clicked

src/rb-select-control/rb-select-control.css
```diff
@@ -5,12 +5,117 @@
...
+}
+
+/**
+ * Default
+ */
+
+.SelectControl-field {
+    -moz-appearance: none;
+    -webkit-appearance: none;
+    appearance: none; /* 1 */
```
chrisvxd: Does post-css not handle `-moz` and `-webkit` properties

ohskylab:
Need a fix for FF, dropping this back to DNM.

src/rb-select-control/demo/demo.tpl.html
ohskylab: Agree, fixed up.

ohskylab:
> Can't help but feel like the dropdown arrow should react when clicked.

@chrisvxd Yes, but let's stick to the component PSDs and WoT. We'll most likely drop this component out and replace with a custom typeahead in the not-to-distant future anyway.

src/rb-select-control/rb-select-control.css
ohskylab: See [referenced comment](https://github.com/rockabox/rbx_ui_components/pull/146/files#diff-cb970ae2b88c8c9485c02e8cf985a31dR28), which links to explanation at http://git.io/vTwoV

ohskylab:
Mozilla fix in.

src/rb-select-control/rb-select-control.css
chrisvxd: @ohskylab my bad.

PR #144, opened 2015-05-22
Rename campaign-Basics to basics fieldset modifier.

k0nG:
TP: https://rockabox.tpondemand.com/entity/9392

PR #149, opened 2015-05-26
Restore campaignBasics fieldset type for backwards compatability

chrisvxd:
Will be removed at next major version bump.

TP https://rockabox.tpondemand.com/entity/9492


ohskylab:
It's worth explicitly mentioning #144 in this thread.

PR #151, opened 2015-05-28
Updated overlay modal demo to just use overlay modal

k0nG:
The overlay modal demo was actually using a modal confirm which had a overlay modal within it. Demo now shows overlay modal in isolation.


ohskylab:
It's literal, but it works.

PR #152, opened 2015-05-28
Add `.GenericForm` styles

ohskylab:
TP: https://rockabox.tpondemand.com/entity/8777

See docs for detailed implementation notes.

Using an `Inner` class is a bit awkward here, but necessary until there's a functioning `rb-icon`.

I think this will do as a hook for now, but should be refactored soon.


k0nG:
On this.

src/rb-generic-form/index.js
```diff
@@ -0,0 +1,19 @@
+define([
+    './rb-generic-form-directive',
+    './rb-generic-form.css'
```
k0nG: Should include `component/rb-fieldset` here and set it as a dependancy of the module.

ohskylab: Can that be picked up on the Angular part?

k0nG: Yeah I guess that's an angular thing. Will let you off.

PR #156, opened 2015-05-28
Add inherit model to datetime control

chrisvxd:
- New property attribute `inherit-model` with binding test
- Renamed `inherit` to `inherit-date-time` as property rename confusing with fixed tests
- Updated demos

TP https://rockabox.tpondemand.com/entity/9562

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,12 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased][unreleased]
 
-### Removed
+### Added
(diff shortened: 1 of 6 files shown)


ohskylab:
Looks like a breaking change so approve but DNM until we have a minor release out without this (unless anyone's desperate for this to go in next release).

chrisvxd:
Tests will currently be failing due to a watcher, but component works in demo. Will revisit on Monday.

k0nG:
LGTM

ohskylab:
:+1:

PR #165, opened 2015-05-29
Refactor check control

jonman:
- `rb-check-control` now renders a single checkbox input.
- `rb-check-control-group` will render multiple checkboxes.
- `options` attribute replaces `ng-model` to define checkbox content.
- `ng-model` will now return an array of currently checked values.

TP: https://rockabox.tpondemand.com/entity/9534


CHANGELOG.md
```diff
@@ -6,9 +6,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased][unreleased]
 
-### Added
+## Changed
```
ohskylab: Should be `###` (I can't believe I'm posting such a pernickety comment and on a Friday evening as well - sorry!)

jonman: Haha, its a fair point! Will update

test/unit/rb-check-control/rb-check-control-group.spec.js
```diff
@@ -0,0 +1,65 @@
...
+                isolatedScope = element.isolateScope();
+            };
+        }));
+
+        describe('checkedOptions', function () {
+            beforeEach(function () {
+                compileTemplate('<rb-check-control-group options="options" ng-model="ngModel">' +
+                    '</rb-check-control-group>');
+            });
+
```
k0nG: Can we check that it renders the right amount of checkboxes here and sets values correctly?

src/rb-check-control/rb-check-control-group-directive.js
```diff
@@ -0,0 +1,63 @@
...
+                ngModel: '=',
+                options: '=',
+                title: '@'
+            },
+            restrict: 'E',
+            replace: true,
+            template: template,
+            link: function ($scope, $element, $attributes) {
+
+                $scope.$watch('options|filter:{checked:true}', function (newOptions) {
```
k0nG: Neat. :+1:

src/rb-check-control/rb-check-control-select-all-directive.js
```diff
@@ -30,12 +30,12 @@ define([
 
         return {
             scope: {
+                enableSelectAll: '@',
```
k0nG: Why does the select all need this?

jonman: Currently the select all directive also renders a title without the checkbox. I could move that out into the group directive instead and use an `ng-if` to toggle which will fix this (bug)[https://rockabox.tpondemand.com/entity/9495]

test/unit/rb-check-control/rb-check-control-group.spec.js
jonman: Will update.

src/rb-check-control/rb-check-control-select-all-directive.js
k0nG: Yeah move that into a `label` tag inside the group directive.

k0nG:
MOG

PR #168, opened 2015-06-01
Update `rb-site` demo

ohskylab:
An updated version of the old docs from style guide.

### DIFF
--- a/src/rb-site/demo/demo.tpl.html
+++ b/src/rb-site/demo/demo.tpl.html
@@ -1,11 +1,24 @@
-<rb-site>
-    <header>
-        Your header might go here...
-    </header>
-    <rb-main class="Main--background Main--center">
-        ... and your main content might go here ...
-    </rb-main>
-    <footer>
-        ... and your footer might go here.
-    </footer>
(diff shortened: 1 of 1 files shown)


src/rb-site/demo/demo.tpl.html
```diff
@@ -1,11 +1,24 @@
...
-    </footer>
-</rb-site>
+<section class="Section">
+    <div class="RawHTML">
+        <h2>
+            rb-site
+        </h2>
+        <p>
+            The <code>rb-site</code> component should wrap all content on a page and is used in conjunction with
+            <code>rb-main</code> to create sticky footer and centred content effects.
```
k0nG: `centered` ?

ohskylab: TIL that US is [centered](http://www.oxforddictionaries.com/definition/american_english/centered), British is [centred](http://www.oxforddictionaries.com/definition/english/centred).

k0nG: Oh yes sorry you're right.

PR #180, opened 2015-06-01
Use `angular-scroll` for `rb-scrollspy`

chrisvxd:
The scrollspy provided by `angular-strap` did not allow for custom elements for scroll observation, and only allowed for observation on `window/document`, making it unusable in modals or overlays.

- Gutted out angular-strap and replaced with angular-scroll
- Pass in the `id` of the element to observe scrolling via `scroll-id` attr.
- Scroll-spy now animates. Sexy.
- Tests fixed
- Demos fixed

See `angular-scroll` here https://github.com/oblador/angular-scroll.

@ohskylab wasn't sure if I should put the base as `2.0.0` or `master` but feel free to edit it. I would like to see this in the next release!

TP https://rockabox.tpondemand.com/entity/9618


ohskylab:
@chrisvxd Excellent, fast work! Yes to `2.0.0`, but the base should be `master` so will need to re-do this as a new PR.

src/rb-scrollspy/demo/demo.tpl.html
```diff
@@ -1,12 +1,31 @@
+<style type="text/css">
+  #container {
```
ohskylab: @chrisvxd I know this is a) just a demo and b) probably pulled from pulled from `angular-scroll` and c) inline, but is it worth making the CSS more SUIT-like?

src/rb-scrollspy/rb-scrollspy.tpl.html
```diff
@@ -1,19 +1,25 @@
...
+                <span>{{category.label}}</span>
             </a>
 
             <ul>
-                <li ng-repeat="section in category.items track by $index">
-                    <a class="Scrollspy-Section" href="#{{::section.anchor}}" bs-scrollspy data-target="#{{::section.anchor}}" data-offset="{{::offset || 0}}">
-                        <span>{{::section.label}}</span>
+                <li ng-repeat="section in category.items">
+                    <a class="Scrollspy-Section"
+                        href="#{{section.anchor}}"
```
k0nG: Think this should be `{{::section.anchor}}`

src/rb-scrollspy/demo/demo.tpl.html
ohskylab: :ok_hand: with this.

PR #183, opened 2015-06-02
Use `angular-scroll` for `rb-scrollspy`

chrisvxd:
The scrollspy provided by `angular-strap` did not allow for custom elements for scroll observation, and only allowed for observation on `window/document`, making it unusable in modals or overlays.

- Gutted out angular-strap and replaced with angular-scroll
- Pass in the `id` of the element to observe scrolling via `scroll-id` attr.
- Scroll-spy now animates. Sexy.
- Tests fixed
- Demos fixed

See `angular-scroll` here https://github.com/oblador/angular-scroll.

@ohskylab wasn't sure if I should put the base as `2.0.0` or `master` but feel free to edit it. I would like to see this in the next release!

TP https://rockabox.tpondemand.com/entity/9618

### DIFF
--- a/src/rb-scrollspy/rb-scrollspy.tpl.html
+++ b/src/rb-scrollspy/rb-scrollspy.tpl.html
@@ -1,18 +1,24 @@
 <div>
-    <ul>
-        <li ng-repeat="category in categories track by $index">
+    <ul du-scroll-container="{{scrollId}}" du-spy-context>
+        <li ng-repeat="category in categories">
             <a class="Scrollspy-Category"
                href="#{{::category.anchor}}"
-               bs-scrollspy
-               data-target="#{{::category.anchor}}"
-               data-offset="{{::offset || 0}}"
-               ng-class="{'Scrollspy-Category--hidden': category.hidden}">
+               du-scrollspy
+               du-smooth-scroll
+               offset="{{offset || 0}}"
+               ng-class="{'Scrollspy-Category--hidden': category.hidden}"
+            >
                 <span>{{::category.label}}</span>
             </a>
 
             <ul>
-                <li ng-repeat="section in category.items track by $index">
-                    <a class="Scrollspy-Section" href="#{{::section.anchor}}" bs-scrollspy data-target="#{{::section.anchor}}" data-offset="{{::offset || 0}}">
+                <li ng-repeat="section in category.items">
+                    <a class="Scrollspy-Section"
+                        href="#{{::section.anchor}}"
+                        offset="{{offset || 0}}"
+                        du-scrollspy
+                        du-smooth-scroll
+                    >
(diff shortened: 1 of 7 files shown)


chrisvxd:
Recreated as original PR had wrong base (`2.0.0`). Need to update changelog

src/rb-scrollspy/rb-scrollspy.tpl.html
```diff
@@ -1,19 +1,25 @@
...
+                <span>{{category.label}}</span>
             </a>
 
             <ul>
-                <li ng-repeat="section in category.items track by $index">
-                    <a class="Scrollspy-Section" href="#{{::section.anchor}}" bs-scrollspy data-target="#{{::section.anchor}}" data-offset="{{::offset || 0}}">
-                        <span>{{::section.label}}</span>
+                <li ng-repeat="section in category.items">
+                    <a class="Scrollspy-Section"
+                        href="#{{section.anchor}}"
```
k0nG: {{ ::section.anchro }}

src/rb-scrollspy/rb-scrollspy.tpl.html
```diff
@@ -1,19 +1,25 @@
...
-               data-offset="{{::offset || 0}}"
-               ng-class="{'Scrollspy-Category--hidden': category.hidden}">
-                <span>{{::category.label}}</span>
+               href="#{{category.anchor}}"
+               du-scrollspy
+               du-smooth-scroll
+               offset="{{offset || 0}}"
+               ng-class="{'Scrollspy-Category--hidden': category.hidden}"
+            >
+                <span>{{category.label}}</span>
```
k0nG: {{ ::category.label }}

src/rb-scrollspy/rb-scrollspy.tpl.html
```diff
@@ -1,19 +1,25 @@
...
-                    <a class="Scrollspy-Section" href="#{{::section.anchor}}" bs-scrollspy data-target="#{{::section.anchor}}" data-offset="{{::offset || 0}}">
-                        <span>{{::section.label}}</span>
+                <li ng-repeat="section in category.items">
+                    <a class="Scrollspy-Section"
+                        href="#{{section.anchor}}"
+                        offset="{{offset || 0}}"
+                        du-scrollspy
+                        du-smooth-scroll
+                    >
+                        <span>{{section.label}}</span>
```
k0nG: {{ ::section.label }}

chrisvxd:
@k0nG I probably should have put those in a separate commit, but they were removed deliberately as @darwinlopez requested after having issues integrating with CM.

k0nG:
Ok if there is a reason for keeping them as not one time binding's that's fine. I guess the scrollspy does need to re render on searches.

PR #186, opened 2015-06-02
Add `-buttons` descendent to `rb-generic-form`

ohskylab:
TP: https://rockabox.tpondemand.com/entity/9563

Allows any (reasonable) number of button items at the end of a generic form.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- Nothing.
+- [`rb-generic-form` `-buttons` descendent](https://github.com/rockabox/rbx_ui_components/pull/186)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,7 +8,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 
-- Nothing.
+- [`rb-generic-form` `-buttons` descendent](https://github.com/rockabox/rbx_ui_components/pull/186)
```
chrisvxd: Do we not want a description on this? And what's the deal with changelog and Suit-only components?

ohskylab: Thought about adding more description, figured it was fairly clear from context, but this line in the change log should probably be swapped out for an Angular component one, when that's been done. Maybe consider this line a placeholder for now.

PR #192, opened 2015-06-03
Style `rb-definition-list` component

ohskylab:
TP: https://rockabox.tpondemand.com/entity/9376

* Use an unordered list with headings, rather than definition list. Using a definition list element confers no benefits and reduces flexibility.
* Add (optional) toolbar descendant, with (optional) button descendant: mirrors `rb-table` structure.

### DIFF
--- a/src/rb-definition-list/rb-definition-list.tpl.html
+++ b/src/rb-definition-list/rb-definition-list.tpl.html
@@ -1,7 +1,7 @@
 <div class="DefinitionList">
     <div class="DefinitionList-toolbar">
         <div class="DataActionTable-action">
-            <rb-button>Edit Targeting</rb-button>
+            <rb-button ng-if="buttonText" ng-click="buttonClick()">{{ ::buttonText }}</rb-button>
         </div>
     </div>
     <ul class="DefinitionList-main" ng-transclude></ul>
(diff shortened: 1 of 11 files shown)


src/rb-definition-list/rb-definition-list.tpl.html
```diff
@@ -1,2 +1,8 @@
-<dl class="multicolumn" ng-transclude>
-</dl>
+<div class="DefinitionList">
+    <div class="DefinitionList-toolbar">
+        <div class="DataActionTable-action">
+            <rb-button>Edit Targeting</rb-button>
```
nwhite89: What's targeting?

ohskylab: To be fixed by @k0nG!

ohskylab:
Also, this should really be a `<table>`.

ohskylab:
Also, I need to fix the tests.

ohskylab:
Tests fixed at 62a9451997d9cb5a9fb98f92c85be2adf400e0cb. Thanks, @k0nG.

src/rb-definition-list/rb-definition-list.tpl.html
ohskylab: @nwhite89 fixed by @k0nG.

PR #200, opened 2015-06-04
Fix definition list layout

ohskylab:
Long text in the body part was pushing container into title part. Setting `min-width` prevents collapse.


k0nG:
Checked in CM. Looks good.

PR #205, opened 2015-06-08
Disable time modifier for rb-datetime-control component.

jonman:
TP: https://rockabox.tpondemand.com/entity/9217


chrisvxd:
This attribute breaks the `is-required` attribute from correctly setting the form validity.

I.e. if you have a required datetime with `disable-time`, populate the date field, the form will still be invalid.

src/rb-datetime-control/rb-datetime-control.tpl.html
```diff
@@ -20,11 +20,12 @@
         type="text"
         class="TextControl-field"
         name="{{timeName}}"
+        ng-class="{'is-invalid': form[timeName].$touched && form[timeName].$invalid}"
+        ng-disabled="disabledTextInputs()"
+        ng-if="!disableTime"
```
chrisvxd: This doesn't work if we use `disableTime="false"` on the directive, since it is a string here. To fix, use `ng-if="{{disableTime === false}}"`

chrisvxd:
@jonman all good just needs rebase!

PR #234, opened 2015-06-15
Updated `rb-text-control` demo with components for layout

chrisvxd:
Needed for: https://github.com/rockabox/rbx_ui_components/pull/233


ohskylab:
:+1: thanks for caring!

PR #235, opened 2015-06-15
Style `rb-check-with-text-control-group`

ohskylab:
TP: https://rockabox.tpondemand.com/entity/9307


chrisvxd:
Can highlight text on checkbox label, otherwise looks great :+1:

ohskylab:
Yeah, need to fix that in `check-control`, but one for the future...

PR #236, opened 2015-06-15
`rb-grid` `gutter` and `flex-cells` attrs

chrisvxd:
Includes tests and new demo sections.

TP https://rockabox.tpondemand.com/entity/9831


ohskylab:
`flexCells` as separate PR?

chrisvxd:
Added `flex-cells` attribute to this PR

PR #242, opened 2015-06-18
Fix save button disabled binding expression

chrisvxd:
This enables it to watch for changes, making it compatible with CM.


nwhite89:
MoG

PR #255, opened 2015-06-26
`rb-section` should use heading as attribute, not title

ohskylab:
TP: https://rockabox.tpondemand.com/entity/9755

Stops standard HTML `title` attribute tooltips appearing in component.

Breaking change: merge at the same time as rockabox/rbx_campaign_management/pull/155


ohskylab:
Need to update component demos.

PR #271, opened 2015-07-06
Angular fieldset actions modifier

jonman:
- `<rb-fieldset type="actions"></rb-fieldset>` will allow the use of field set items via `transclude`.
- `rb-fieldset-item` subcomponent directive.

TP: https://rockabox.tpondemand.com/entity/10097

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 - [`rb-deep-search` component](https://github.com/rockabox/rbx_ui_components/pull/232)
 - ['rb-scrollspy' do not show hidden sections](https://github.com/rockabox/rbx_ui_components/pull/232)
 - ['rb-check-control-select-all' disabled if all child options are disabled](https://github.com/rockabox/rbx_ui_components/pull/235)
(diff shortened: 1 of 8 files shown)


src/rb-fieldset/rb-fieldset-directive.js
```diff
@@ -23,8 +23,12 @@ define([
...
         return {
             scope: {
-                type: '@',
-                label: '@'
+                label: '@',
+                primaryAction: '&?',
+                primaryLabel: '@',
+                secondaryAction: '&?',
+                secondaryLabel: '@',
+                type: '@'
```
chrisvxd: This seems overly complex as using subs, @ohskylab overheard you discussing this earlier today... any further thoughts?

chrisvxd:
@jonman awesome commit history!!

src/rb-fieldset/rb-fieldset-directive.js
ohskylab: Our conclusion (forgive me for paraphrasing, @jonman):

1. Only needs to do what we can see right now (two actions, both are buttons)
2. Use the same approach as with `rb-action-bar`, as they're very similar components

If that's the case, :+1: from me.

PR #279, opened 2015-07-10
Angularify `rb-generic-form` component

ohskylab:
TP: https://rockabox.tpondemand.com/entity/9549

Note that this component currently has no states: the "Required Field" message is always displayed, which is ultimately all this component is responsible for.


jonman:
@ohskylab would mind updating your description to reflect the addition of tests?

PR #293, opened 2015-07-13
Use real components in `rb-modal-confirm` demo

ohskylab:
No TP.

### DIFF
--- a/src/rb-modal-confirm/demo/demo.tpl.html
+++ b/src/rb-modal-confirm/demo/demo.tpl.html
@@ -1,33 +1,43 @@
-<div class="ComponentView">
-
-    <h2>ModalConfirm with default button labels</h2>
-    <rb-button ng-click="demoCtrl.showModal1 = true">Show Modal 1</rb-button>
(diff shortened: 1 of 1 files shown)


src/rb-modal-confirm/demo/demo.tpl.html
```diff
@@ -1,33 +1,43 @@
...
+        <rb-page-title heading="rb-modal-confirm">
+        </rb-page-title>
+        <rb-section heading="Default Button Labels">
+            <div class="ComponentView">
+                <rb-button ng-click="demoCtrl.showModal1 = true">Show Modal 1</rb-button>
+            </div>
+        </rb-section>
+        <rb-section heading="Custom Button Labels">
+            <div class="ComponentView">
+                <rb-button ng-click="demoCtrl.showModal2 = true">Show Modal 1</rb-button>
```
jonman: `Show Modal 2`

ohskylab: Good spot! I'm tackling easy doc fixes as part of this process, so added an additional commit for that.

PR #306, opened 2015-07-14
Add `raw-html-display` component

ohskylab:
TP: https://rockabox.tpondemand.com/entity/10403

Review by commit for great sanity.

Please merge rockabox/rbx_style_guide#197 shortly after this, to prevent doubling-up of styles.

The original `.RawHTML` styles weren't used in CM, so no fix needed there.


ohskylab:
Needs tests.

ohskylab:
Doesn't need tests (yet).

PR #307, opened 2015-07-14
rb-generic-form has a 'hasRequiredFields' switch

jess0407:
- now can pass attr ‘has-required-fields’
- written the corresponding tests
- updated the component demo

https://rockabox.tpondemand.com/entity/10440

### DIFF
--- a/src/rb-generic-form/demo/demo.tpl.html
+++ b/src/rb-generic-form/demo/demo.tpl.html
@@ -21,5 +21,19 @@
                 </rb-generic-form>
             </div>
         </rb-section>
+        <rb-section heading="Has required fields.">
+            <div class="ComponentView">
+                <rb-generic-form has-required-fields="true">
+                    Form content.
+                </rb-generic-form>
+            </div>
+        </rb-section>
+        <rb-section heading="Does not have required fields.">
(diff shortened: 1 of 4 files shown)


src/rb-generic-form/demo/demo.tpl.html
```diff
@@ -21,5 +21,19 @@
                 </rb-generic-form>
             </div>
         </rb-section>
+        <rb-section heading="Has required fields.">
+            <div class="ComponentView">
+                <rb-generic-form has-required-fields='true'>
```
nwhite89: would you mind changing this to double quotes please same as the false, generally HTML should be written with double quotes but we haven't managed to integrate a decent HTML Linter as of yet.

nwhite89:
other than that @ohskylab can you review look wise, the code is fine by me but if you could sign off it does what you need.

ohskylab:
LGTM. Will need a squash. @jess0407, give me a shout if you want a hand with doing a fixup commit.

src/rb-generic-form/demo/demo.tpl.html
jess0407: Noted and changed, thanks Nick.

PR #310, opened 2015-07-15
Set the default debounce time for text controls to ZERO

darwinlopez:
NO TP

Quick change to set the default debounce time for text controls to ZERO


ohskylab:
@darwinlopez: think it's worth editing your commit message to say why this change is necessary?

ohskylab:
@rockabox/frontend do we still need this change, or can we burn it?

PR #311, opened 2015-07-15
Fix `rb-grid` demo

ohskylab:
TP: https://rockabox.tpondemand.com/entity/10476

Unclosed `rb-raw-html-display` broke layout.


ohskylab:
DNM: want to avoid costly rebase on another PR.

PR #316, opened 2015-07-16
Create new badge for status PendingLive

jess0407:
https://rockabox.tpondemand.com/entity/10493

- update the directive and template
- update the demo
- write the test


src/rb-badge/rb-badge.css
```diff
@@ -127,6 +127,24 @@
 }
 
 /**
+ * Status PendingLive modifier
+ */
+
+:root {
+    --Badge--statusPending-base: var(--color-status-pending-base);
```
ohskylab: We're redefining this custom property here. Should be:

```
--Badge--statusPendingLive-base: var(--color-status-pending-live-base);
```

The rest of the `PendingLive` styles should then reference this.

src/rb-badge/rb-badge.css
```diff
@@ -127,6 +127,24 @@
...
+}
+
+.Badge--statusPendingLive .Badge-body {
+    background: var(--Badge--statusPending-base) url("./images/Badge--statusPending.svg") no-repeat center center;
```
ohskylab: It makes no practical difference but for the sake of consistency within this component, copy `Badge--statusPending.svg` -> `Badge--statusPendingLive.svg` and change background refs.

jess0407: All updated :) thx

PR #318, opened 2015-07-17
Update the Scheduled status

jess0407:
- update the directive and template
- update the demo
- update the test

https://rockabox.tpondemand.com/entity/9186


jess0407:
This has been updated per the new status 'Scheduled' as well

PR #319, opened 2015-07-20
Set ui-sref-opts's reload option to true.

jess0407:
https://rockabox.tpondemand.com/entity/10499


src/rb-page-title/rb-page-title.tpl.html
```diff
@@ -1,9 +1,10 @@
+ui-sref
```
nwhite89: ???

nwhite89:
@jess0407 this needs to go to develop not master

PR #320, opened 2015-07-23
New Validation on too many digits

jess0407:
Part of 
https://rockabox.tpondemand.com/entity/10634

### DIFF
--- a/src/rb-text-control/custom-validation-directive.js
+++ b/src/rb-text-control/custom-validation-directive.js
@@ -28,9 +28,22 @@ define([
                             splitValue = (viewValue) ? viewValue.split('.') : [],
                             decimalsPlaces = (splitValue[1]) ? splitValue[1].length : 0,
                             result = ctrl.$isEmpty(modelValue) || (decimalsPlaces <= maxDecimals);
-
                         return result;
                     };
+
+                    ctrl.$validators.tooManyDigits = function (modelValue, viewValue) {
+                        var wholeNumberDigits, result;
+                        if (!modelValue) {
+                            wholeNumberDigits = 0;
+                        } else if (modelValue.toString().split('e+').length > 1) {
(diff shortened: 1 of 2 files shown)


src/rb-text-control/custom-validation-directive.js
```diff
@@ -28,9 +28,17 @@ define([
                             splitValue = (viewValue) ? viewValue.split('.') : [],
                             decimalsPlaces = (splitValue[1]) ? splitValue[1].length : 0,
                             result = ctrl.$isEmpty(modelValue) || (decimalsPlaces <= maxDecimals);
+                        return result;
+                    };
 
+                    ctrl.$validators.tooManyDigits = function (modelValue, viewValue) {
+                        var wholeNumberDigits = !modelValue ? 0 :
```
nwhite89: Could we split this out a bit to use if statements? quite difficult to follow as an inline if statement looks like there is two in one?

src/rb-text-control/custom-validation-directive.js
```diff
@@ -28,9 +28,22 @@ define([
...
+                    ctrl.$validators.tooManyDigits = function (modelValue, viewValue) {
+                        var wholeNumberDigits;
+                        if (!modelValue) {
+                            wholeNumberDigits = 0;
+                        } else if (modelValue.toString().split('e+').length > 1) {
+                            wholeNumberDigits = modelValue.toString().split('e+')[1];
+                        } else {
+                            wholeNumberDigits = modelValue.toString().split('.')[0].length;
+                        }
+                        result = ctrl.$isEmpty(modelValue) || (wholeNumberDigits <= 18);
```
nwhite89: result isn't defined here? as a variable possibly just return instead of assigning? up to you

jess0407:
screen shot 2015-07-23 at 12 23 06
just keep consistency with other validators :)

PR #323, opened 2015-07-27
Add `rb-divider` component

ohskylab:
TP: https://rockabox.tpondemand.com/entity/10735


ohskylab:
Fixups squashed, MoG.

PR #325, opened 2015-07-27
Add `rb-editor-wrapper` component

ohskylab:
TP: https://rockabox.tpondemand.com/entity/10463

Needs additional Angular work outside this PR.


chrisvxd:
Are you planning on carrying out the angular work? It is just a serious of transcluded directives. Would make sense to do it here IMO.

src/rb-editor-wrapper/rb-editor-wrapper.tpl.html
```diff
@@ -0,0 +1,13 @@
...
+            header / page-header
+        </div>
+        <div class="EditorWrapper-inner">
+            <div class="EditorWrapper-canvas">
+                canvas
+            </div>
+            <div class="EditorWrapper-form">
+                editor forms
+            </div>
+        </div>
```
chrisvxd: Did you manage to speak to @annahamzah regarding footer?

ohskylab: No - skipped that. It's not in wires or designs and we're aiming for demo, so following process.

ohskylab:
> Are you planning on carrying out the angular work? It is just a serious of transcluded directives. Would make sense to do it here IMO.

@chrisvxd I want feedback on whether this approach is going to work, naming etc before splitting it out - it's easier to work with statics if changes are necessary. Also means someone else could pick up the Angular task if necessary.

jonman:
I think it is also better to keep these static for now as there will be complexities around multi-transclude and incorporating ui-view into the directives.

PR #326, opened 2015-07-27
Fix rb-ratio math fail

jess0407:
Fix math fail - denominator cannot be zero

https://rockabox.tpondemand.com/entity/9393


nwhite89:
@jess0407 could you squash these commits please

PR #327, opened 2015-07-27
Add static `rb-tabs` component

ohskylab:
TP: https://rockabox.tpondemand.com/entity/10642

Needs Angular work.

### DIFF
--- a/src/rb-tabs/demo/demo.tpl.html
+++ b/src/rb-tabs/demo/demo.tpl.html
@@ -1,0 +1,19 @@
+<rb-site>
+    <rb-main>
+        <rb-page-title heading="rb-tabs">
+        </rb-page-title>
+        <rb-section heading="Introduction" hide-heading="true">
+            <rb-raw-html-display>
+                <p>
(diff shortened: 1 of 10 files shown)


src/rb-tabs/demo/demo.tpl.html
```diff
@@ -0,0 +1,19 @@
...
+        </rb-page-title>
+        <rb-section heading="Introduction" hide-heading="true">
+            <rb-raw-html-display>
+                <p>
+                    The <code>rb-tabs</code> component...
+                </p>
+            </rb-raw-html-display>
+        </rb-section>
+        <rb-section heading="Default">
+            <!-- <rb-component-display> -->
```
ohskylab: Commented because the `.is-active` background colour is the same as the one used for `rb-component-display`, which makes it hard to see what's going on. We should probably change `rb-component-display` to use a background colour that isn't from our current palette to avoid this.

jess0407: Sounds like a good idea to me

PR #335, opened 2015-07-29
Angularfy tabset and tab

jess0407:
https://rockabox.tpondemand.com/entity/10643


src/rb-tabs/index.js
```diff
@@ -1,7 +1,8 @@
 define([
-    './rb-tabs-directive',
+    './tabset-directive',
+    './tab-directive',
     './rb-tabs.css'
-], function (rbTabsDirective, css) {
+], function (tabset, tab, css) {
```
nwhite89: can you please make sure you keep to the naming convention here should be ``rbTabsetDirective``

src/rb-tabs/demo/demo.tpl.html
```diff
@@ -11,9 +11,24 @@
         </rb-section>
         <rb-section heading="Default">
             <!-- <rb-component-display> -->
-                <rb-tabs></rb-tabs>
+            <tabset>
+                <tab heading="Tab 1">
```
nwhite89: `rb-tab`

src/rb-tabs/demo/demo.tpl.html
```diff
@@ -11,9 +11,24 @@
         </rb-section>
         <rb-section heading="Default">
             <!-- <rb-component-display> -->
-                <rb-tabs></rb-tabs>
+            <tabset>
```
nwhite89: `rb-tabset`

nwhite89:
I think in general there needs to be some more unit tests around this to ensure it does what is needed

jess0407:
This is a replica of the existing AngularJS tab component out there, so basically it is how angular-ui people implement it. And they use the `href=''` and binding with the heading. To me, either way don't make a difference (as it turns out both ways the code works) so I have amended the code per suggestion.

jess0407:
@nwhite89 I have also added some unit tests including tests for `addTab` and `select` function, let me know if this is ready to merge.

PR #338, opened 2015-07-29
Fix `panel-scroll` `unscrollable` Firefox bug

ohskylab:
TP: https://rockabox.tpondemand.com/entity/10811

Only affects `rb-scrollspy` at time of fix (so test using `rb-scrollspy` either in ui_components or in an application).

Setting a `min-height: 0;` on the `unscrollable` directive fixes broken scrolling of child components in Firefox.

See: http://stackoverflow.com/a/26916542/2270732

### DIFF
--- a/src/rb-panel-content/rb-panel-content.css
+++ b/src/rb-panel-content/rb-panel-content.css
@@ -6,6 +6,8 @@
 
 /**
  * 1. Set to viewport height (we can do this as overlay panel uses full viewport height).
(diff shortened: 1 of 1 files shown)


chrisvxd:
Tried to review but can't reproduce bug so can't tell if fixed o.O

ohskylab:
With Firefox and on stage, try scrolling in any targeting panel on stage that uses `rb-scrollspy`. If you can see there, you should see the same problem in the `rb-scrollspy` demo.

PR #342, opened 2015-07-31
Refactor rb-page-header to use background images

ohskylab:
TP: https://rockabox.tpondemand.com/entity/10914

Merge at the same time as:

* rockabox/rbx_campaign_management#224
* rockabox/rig_create#24


ohskylab:
Needs matching PRs for create and distribute.

PR #344, opened 2015-08-04
small css to fix input components border colour on initialisation

gregbenner:
https://rockabox.tpondemand.com/entity/10696


src/rb-text-control/rb-text-control.css
```diff
@@ -114,6 +114,7 @@
     background: var(--TextControl-field-background-color) url(var(--TextControl-field-required-image)) no-repeat;
     background-position: var(--TextControl-field-required-image-position); /* 4 */
     outline: none; /* 2 */
+    box-shadow: none;
```
jaylett: Since this fixes a bug in a specific rendering engine, it'd be nice to have a numbered comment reference, as with other rules here.

jaylett:
Non-functional suggestion: the commit message isn't going to be particularly helpful in the far future. We have some [loose guidelines](https://docs.rockabox.com/rbx3/new/process/style-guide/git-commits.html#git-commits-style-guide) on commits which may help.

src/rb-text-control/rb-text-control.css
ohskylab: :+1: It might also be worth filtering to only target Gecko, as well, possibly via the [:-moz-ui-invalid](https://developer.mozilla.org/en-US/docs/Web/CSS/:-moz-ui-invalid) pseudo-class.

I'm going to some quick tests with some proper forms - will update when done.

ohskylab: I'd be happier with this, after the `.TextControl-field`:

```
.TextControl-field:-moz-ui-invalid {
    box-shadow:none;
}
```
It's a pseudo-class, so uses a [single `:`](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements#Notes).

gregbenner:
TP https://rockabox.tpondemand.com/entity/10696

ohskylab:
@gregbenner Fix looks good but commits need squashing before merge: please ask me if you need a hand with this.

gregbenner:
Ok I think it's all squished now?

PR #345, opened 2015-08-05
Fix Inherit from campaign gives invalid date format entered

jess0407:
Also fixed the some demo does not validate datetime

https://rockabox.tpondemand.com/entity/6208


nwhite89:
@jess0407 do you think it's possible to get a regression test for this ?

jess0407:
@nwhite89  let me take a look

PR #346, opened 2015-08-06
Add background image modifiers to `rb-definition-list`

ohskylab:
TP: https://rockabox.tpondemand.com/entity/10933

In place of `rb-icon`.

Needs additional Angularification to set class modifier:

```html
<div class="DefinitionList-icon DefinitionList-icon--modifierName">
```


ohskylab:
Needs a tiny fix: remove `DefinitionList-icon--segmentation` in view.

PR #351, opened 2015-08-10
fix browser behavior for firefox that reports invalid inputs before u…

gregbenner:
fix browser behavior for firefox that reports invalid inputs before user input


ohskylab:
Rejigger of #344.

PR #353, opened 2015-08-10
fix debounce on password input fields to 0 instead of 700. Moves ng-model-options to a controller for better control and logic.

gregbenner:
fix debounce on password input fields to 0 instead of 700. Moves ng-model-options to a controller for better control and logic. Debounce is not needed on password field since it is not sending server requests.


nwhite89:
@darwinlopez was looking at something like this before Darwin could you take a look over this

PR #355, opened 2015-08-10
Jessie/11028 rg text control

jess0407:
https://rockabox.tpondemand.com/entity/11028
* rg-text-control for `url` and `text`
* unit testing for rg-text-control


src/rg-text-control/rg-text-control.css
```diff
@@ -0,0 +1,220 @@
...
+.RgTextControl--currency .RgTextControl-field:focus:required:invalid {
+    background-image: url(var(--RgTextControl--currency-field-invalid-image)), url(var(--RgTextControl-field-required-invalid-image));
+    background-position: var(--RgTextControl-field-type-image-position), var(--RgTextControl-field-required-image-position);
+}
+
+.RgTextControl--currency .RgTextControl-field:disabled,
+.RgTextControl--currency .RgTextControl-field.is-disabled {
+    background-image: url(var(--RgTextControl--currency-field-image)), url(var(--RgTextControl-field-disabled-image)) !important;
+    background-position: var(--RgTextControl-field-type-image-position), var(--RgTextControl-field-disabled-image-position) !important;
+}
```
chrisvxd: Are most of these moved directly over from the original `rb-text-control`?

src/rg-text-control/rg-text-control.css
```diff
@@ -0,0 +1,220 @@
...
+    background-position: var(--RgTextControl-field-type-image-position), var(--RgTextControl-field-required-image-position);
+}
+
+.RgTextControl--currency .RgTextControl-field:required:focus,
+.RgTextControl--currency .RgTextControl-field.is-required:focus,
+.RgTextControl--currency .RgTextControl-field.is-populated:required,
+.RgTextControl--currency .RgTextControl-field.is-populated.is-required {
+    background-image: url(var(--RgTextControl--currency-field-focus-image)), url(var(--RgTextControl-field-required-image));
+    background-position: var(--RgTextControl-field-type-image-position), var(--RgTextControl-field-required-image-position);
+}
```
chrisvxd: Interesting how we have to now use angular classes to handle these states

src/rg-text-control/rg-text-control.css
jess0407: @chrisvxd  yes, the css file is a exact replica with `Rg` in front of `TextControl` and I believe a width statement addition.

src/rg-text-control/rg-text-control.tpl.html
```diff
@@ -0,0 +1,46 @@
...
+    'RgTextControl--number': type==='number',
+    'RgTextControl--search': type==='search',
+    'RgTextControl--textarea': type==='textarea',
+    'RgTextControl--currency': type==='currency',
+    'RgTextControl--url': type==='url',
+    'RgTextControl--small': small==='true'
+}">
+    <label
+        class="RgTextControl-label u-textLabel"
+        ng-class="{'is-invalid': form[name].$touched && form[name].$invalid}">
```
chrisvxd: I'm not sure how @ohskylab will feel about ditching SUIT conventions for generic modifiers like this... should probably be `RgTextControl--invalid` or similar.

src/rg-text-control/rg-text-control.css
chrisvxd: Are we using `is-required` etc in `TextControl`?

src/rg-text-control/rg-text-control.tpl.html
jess0407: I collaborated with @ohskylab on this. We did a quick find and replace to separate the new classes from the old text-control classes. Not sure about the SUIT convention, maybe we should ask Paul tomorrow.

chrisvxd:
As discussed with @jess0407, waiting on thoughts from @ohskylab regarding non-SUITy nature of modifiers

src/rg-text-control/rg-text-control.tpl.html
ohskylab: Modifiers are for configuration (e.g. `.RgTextControl--url`), is-stateName classes are for changes to the state of a component (e.g. `.RgTextControl.is-invalid`). See docs on [is-stateOfComponent](https://github.com/suitcss/suit/blob/master/doc/naming-conventions.md#is-stateOfComponent).

This is the same as we have in `rb-text-control`, so should probably wave it on through in this PR regardless :)

chrisvxd: @ohskylab aha. I always assumed states were modifiers. Not sure where I got that from.

PR #356, opened 2015-08-10
Add version lock styles to `rb-canvas`

ohskylab:
TP: https://rockabox.tpondemand.com/entity/10994

Default state is "Locked Mode":

```html
<div class="Canvas">
    <div class="Canvas-state">
        <div class="Canvas-stateInner">
            Locked Mode
        </div>
    </div>
    <div class="Canvas-inner" ng-transclude>
    </div>
</div>
```

`Canvas--edit` modifier sets "Edit Mode":

```html
<div class="Canvas Canvas--edit">
    <div class="Canvas-state">
        <div class="Canvas-stateInner">
            Edit Mode
        </div>
    </div>
    <div class="Canvas-inner" ng-transclude>
    </div>
</div>
```


chrisvxd:
No demo for the `edit` modifier?

ohskylab:
I was going to leave it for the Angular implementation as this component is already in use, but you're right, I think I may hardcode a demo, to be fixed up during Angularification.

ohskylab:
@chrisvxd I've updated the demo and added a tiny additional commit to prevent the label leaking out when no creative present.

chrisvxd:
@ohskylab "in review" is for when someone is actively reviewing it! We should probably document them somewhere.

PR #378, opened 2015-08-18
Story/10920 refactor rb side nav to use background images

gregbenner:
Refactor rb-side-nav to use background images
TP: https://rockabox.tpondemand.com/entity/10920


ohskylab:
Two things...

1. Needs rebasing
2. We should have an equivalent PR in CM to update this component

PR #384, opened 2015-08-21
Refactor rb-page-header into a new component rg-header

ohskylab:
TP: https://rockabox.tpondemand.com/entity/11030

Refactor `rb-page-header` into a new component, `rg-header`, to support colour variants and be more componenty.

Leave the existing `rb-page-header` and related components as-is for now.

TODO:

* [x] Fix `rb-user` icon hover colour
* [x] Move to `rg-header` (will retire `rb-header` later)
* [x] Confirm `.is-selected` state exists and works
* [x] Refactor per planning
* [x] Fold in `rb-nav-bar` styles
* [x] Set up Angular colour classes, tidy demo


chrisvxd:
Waiting on rename of `rg-page-header` to `rg-header` and a squash before merge.

ohskylab:
Back to DNM: `is-selected` class issue needs resolving.

chrisvxd:
@ohskylab DNM readied until note added to old `rb-header` demo, pointing to new component as discussed

PR #385, opened 2015-08-24
Fix login button delayed to activate bug.

jess0407:
Use rg-text-control in the form
https://rockabox.tpondemand.com/entity/10711

### DIFF
--- a/src/rb-login-form/rb-login-form.tpl.html
+++ b/src/rb-login-form/rb-login-form.tpl.html
@@ -5,26 +5,27 @@
     <form name="loginForm" class="LoginForm-body">
         <fieldset>
             <div class="LoginForm-control">
-                <rb-text-control
+                <rg-text-control
                     form="loginForm"
-                    label="Email"
-                    placeholder="Email"
-                    name="email"
-                    type="email"
-                    ng-model="emailModel"
-                    is-required="true">
-                </rb-text-control>
+                    label="Email">
+                    <input type="email"
+                           name="email"
+                           placeholder="Email"
+                           ng-model="emailModel"
+                           ng-required="true"
+                           custom-validation>
+                </rg-text-control>
             </div>
             <div class="LoginForm-control">
-                <rb-text-control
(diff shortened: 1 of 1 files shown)


src/rb-login-form/rb-login-form.tpl.html
```diff
@@ -5,26 +5,26 @@
...
                     form="loginForm"
-                    label="Email"
-                    placeholder="Email"
-                    name="email"
-                    type="email"
-                    ng-model="emailModel"
-                    is-required="true">
-                </rb-text-control>
+                    label="Email">
+                    <input type="email"
```
jonman: Unfortunately, the email type will need our custom validator that we have in the old `rb-text-control`.

https://github.com/rockabox/rbx_ui_components/blob/master/src/rb-text-control/custom-validation-directive.js#L4

jess0407: @jonman why is that?

jonman: The issue is similar to the `type="ur"l` default angular validator where `type="email"` thinks that `me@example` is as a valid email address.

jess0407: @jonman cheers noted. I think this pr can be closed though, as I will address adding custom validators on type 'url' and 'email'  for rg-text-control on a separate pr addressing tp ticket https://rockabox.tpondemand.com/entity/11309

ohskylab: Would we want to merge this to `master` without having custom validators in place? Any risk it'll create regressions in CM, even if just until this is picked up in the other story?

nwhite89: agreed, there is a PR doing this I have seen open I've made this as Do not merge and like to see the other pr merged and finished first #391

ohskylab:
OK to merge now #391 is through?

nwhite89:
@jess0407 can you rebase please

PR #386, opened 2015-08-24
Add static `rb-device-choice-control` component

ohskylab:
TP: https://rockabox.tpondemand.com/entity/9303


ohskylab:
DNM because we might abandon this approach in favour of extending `rb-check-control-group`.

ohskylab:
Closing, we have a new approach.

PR #387, opened 2015-08-24
Refactor rb side nav to use background images

gregbenner:
Refactor rb-side-nav to use background images instead of rb-badge


gregbenner:
@ohskylab can you have a peak..?

test/unit/rb-side-nav/rb-side-nav-item.spec.js
```diff
@@ -108,18 +108,6 @@ define([
             });
         });
 
-        describe('icon attribute', function () {
-
-            it('should use icon form icon attribute', function () {
-                compileTemplate('<rb-side-nav-item icon="blue-base-16-geo"></rb-side-nav-item>');
```
nwhite89: maybe instead of deleting this amend to ensure that it adds the correct class that it needs now?

gregbenner: Ah yeah can write a test, think @ohskylab removed this to pass the build step prior.

nwhite89: ok

ohskylab: Figured that test wasn't needed any more :)

Yes, should probably add one for new structure. @gregbenner, are you OK with adding?

ohskylab:
LGTM. I think we need a matching PR in CM, otherwise this will break stuff when merged...

test/unit/rb-side-nav/rb-side-nav-item.spec.js
gregbenner: Yep no problemo, adding now.

gregbenner:
Pairs well with https://github.com/rockabox/rbx_campaign_management/pull/260

ohskylab:
Historians: mentions of `rb-badge` should be `rb-icon`.

gregbenner:
Ready to be accepted? :cactus:

gregbenner:
#angular #reviewme

PR #388, opened 2015-08-25
Add is-device-control switch on rb-check-group-control

jess0407:
https://rockabox.tpondemand.com/entity/9303

merge with 
https://github.com/rockabox/rbx_campaign_management/pull/261


ohskylab:
@jess0407 Needs to add modifier to root element when `isDeviceChoice`, like this:

```html
<div class="CheckControlGroup CheckControlGroup--deviceChoice">
```
Then I think we're there.

PR #394, opened 2015-08-25
RgHeader-user

chrisvxd:
Needs review from @ohskylab and someone else angular from @rockabox/frontend.

**NB THIS IS MERGING INTO ANOTHER STORY BRANCH, NOT MASTER**

Important parts:
- Copied `rb-user` to `rg-header-user` descendent in new `rg-header` component.
- Refactored to include new wrapper.
- Split out `item` from `rg-header-user` so is composable.

See commit history for full description.

TP https://rockabox.tpondemand.com/entity/11030


nwhite89:
Angular point of view looks alright @ohskylab take a look on the layout side

ohskylab:
LGTM. Agree with moving `User` styles into descendant of `RgHeader`. It was done before knowing the shape of things and doesn't need that level of flexibility.

PR #395, opened 2015-08-25
Angularify RgHeader color classes

chrisvxd:
Used a one-time binding approach via the `link` method on the directive to reduce unnecessary watchers.

A thought: this could be abstracted to generically allow for modifiers on all components via a `modifier` attr.

TP https://rockabox.tpondemand.com/entity/11370


ohskylab:
@chrisvxd Is this DNM only because #394 needed merging first, or was there another reason?

src/rg-header/rg-header-directive.js
```diff
@@ -42,7 +42,13 @@ define([
             restrict: 'E',
             replace: true,
             transclude: true,
-            template: template
+            template: template,
+            link: function (scope, ele, attrs) {
+                scope.modifier = '';
+                if (angular.isDefined(attrs['color'])) {
```
nwhite89: do we want a default colour? @ohskylab ?

nwhite89: if not angular point of view GTM

ohskylab: Thought about it, but reckon it should be explicitly stated due to the way the guidelines work.

There's a neutral, no-modifier fallback colour in styles to cover no-modifier situations. It doesn't look bad but it also doesn't look correct.

nwhite89: cool that's fine thought probably the case but thought better check :+1: :)

PR #397, opened 2015-08-26
Adds a centered modifier to rb-section.

nwhite89:
Using attributes will set whether or not to use the centered modifier when using `rb-section`.

TP: https://rockabox.tpondemand.com/entity/11468


ohskylab:
LGTM :+1:.

PR #400, opened 2015-08-26
Attribute transclusion and is-active attr

chrisvxd:
- Add `is-active` toggle to `rg-header-item`
- `transcludeAttrs` method to move attrs to first child with `ng-transclude`
- Integrate `transcludeAttrs` on `rg-header-item` for `sref` and `href`

TP https://rockabox.tpondemand.com/entity/11496


gregbenner:
LGTM

PR #402, opened 2015-08-27
Map in build to incomplete.

nwhite89:
Allow for mapping states to other state classes (currently only `In Build` will map to `Incomplete`).

TP: https://rockabox.tpondemand.com/entity/11400

### DIFF
--- a/src/rb-badge/rb-badge-directive.js
+++ b/src/rb-badge/rb-badge-directive.js
@@ -9,6 +9,7 @@ define([
     var STATE_DETAILS = {
         'warning': 'Warning(s) on this item',
         'statusIncomplete': 'Incomplete',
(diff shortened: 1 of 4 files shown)


src/rb-badge/rb-badge-directive.js
```diff
@@ -9,6 +9,7 @@ define([
     var STATE_DETAILS = {
         'warning': 'Warning(s) on this item',
         'statusIncomplete': 'Incomplete',
+        'statusIn Build': 'Incomplete',
```
chrisvxd: Wouldn't this be better as `statusInBuild`?

nwhite89: API provides In Build this cannot be changed.

PR #404, opened 2015-08-27
Transclude ui sref and remove logo link

chrisvxd:
- Transclude `ui-sref`, not `sref` in `rg-header-item`
- Remove `ng-click` and `pointer` from `RgHeader-logo` as nowhere clear for it to go

TP https://rockabox.tpondemand.com/entity/11030

### DIFF
--- a/src/rg-header/rg-header.css
+++ b/src/rg-header/rg-header.css
@@ -53,7 +53,6 @@
     background-image: url("./images/logo-rockabox-blue.svg"); /* 1 */
     background-position: center;
     background-repeat: no-repeat;
-    cursor: pointer;
     margin: 0 var(--layout-gutter); /* 2 */
     margin-bottom: 0.5em; /* 3 */
     margin-top: 0.5em; /* 3 */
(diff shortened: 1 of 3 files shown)


src/rg-header/rg-header.css
```diff
@@ -53,7 +53,6 @@
     background-image: url("./images/logo-rockabox-blue.svg"); /* 1 */
     background-position: center;
     background-repeat: no-repeat;
-    cursor: pointer;
```
ohskylab: Would it be wrong and bad to comment out this line, rather than remove? I think it if goes, it'll be hard for someone to remember to put it back when it's a link again.

ohskylab:
LGTM, besides comment. @nwhite89?

src/rg-header/rg-header.css
nwhite89: it is a link why do we not want it :/

ohskylab: I think @chrisvxd was removing the logo link because it's somehow ended up with `demoCtrl` stuff mixed in and there's no obvious place for it to link to now we have more than one service and potentially varying user rights. Think we should discuss with @helloyip - I'll page him.

ohskylab: Historians: we're temporarily removing the logo link because we need to plan where it links to. Possibly to a dashboard?

PR #405, opened 2015-08-27
Apply some flex box magic to radio groups that are horizontal

gregbenner:
Apply some flex box magic to radio groups that are horizontal

### DIFF
--- a/src/rb-radio-control/rb-radio-control.css
+++ b/src/rb-radio-control/rb-radio-control.css
@@ -19,6 +19,7 @@
 .RadioControl--row .RadioControl-items {
     display: flex;
     flex-wrap: wrap;
+    justify-content: space-around;
 }
 
 .RadioControl--row .RadioControl-item {


gregbenner:
@ohskylab when you have a sec.. I visually checked over campaign manager and couldn't spot any negatives as per discussion. ¯\_(ツ)_/¯

ohskylab:
I know this is a bolt-on to the intended purpose of this PR, but I think we should turn this PR into a "Tidy up rb-radio-control", because it's all a bit wrong at the mo.

1. Add demos for the two different rb-radio-control with icon types (should really have done this when updating CM).
2. Use `is-row="true"` on these icon demos, so can test for impact outside CM.
3. Fix docs, specifically "Exactly two radio button and label pairs must be displayed when using the row modifier." 
    * Doesn't hold true for icons, so amend to "...when using the row modifier without the icon modifier."
    * Use only two icons on the no-icon row modifier demo, so demo is accurate.

gregbenner:
Ok closing for now until a ticket appears as it needs more attention.

ohskylab:
Added a story for other fixes, so I'm going to reopen this and merge.

PR #411, opened 2015-09-01
Fix the rb-login-form bug

jess0407:
https://rockabox.tpondemand.com/entity/11557

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -28,9 +28,6 @@
     "angular-ui-router": "~0.2.13",
     "rbx_style_guide": "rockabox/rbx_style_guide#develop"
   },
-  "resolutions": {
-    "angular": "=1.3.16"
-  },
   "dependencies": {
     "angular-aria": "~1.3.0",
     "angular-elastic": "rockabox/angular-elastic#master",
--- a/src/rb-login-form/index.js
+++ b/src/rb-login-form/index.js
@@ -2,10 +2,10 @@ define([
     '../rb-button',
     '../rb-icon',
     '../rb-form-message',
-    '../rb-text-control',
+    '../rg-text-control',
     './rb-login-form-directive',
     './rb-login-form.css'
-], function (rbButton, rbIcon, rbFormMessage, rbTextControl, rbLoginFormDirective, css) {
+], function (rbButton, rbIcon, rbFormMessage, rgTextControl, rbLoginFormDirective, css) {
     /**
      * @ngdoc module
      * @name rb-login-form
@@ -19,7 +19,7 @@ define([
             rbButton.name,
             rbIcon.name,
             rbFormMessage.name,
-            rbTextControl.name
+            rgTextControl.name
         ])
         .directive('rbLoginForm', rbLoginFormDirective);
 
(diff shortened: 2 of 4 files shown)


bower.json
```diff
@@ -19,7 +19,7 @@
     "**/.*"
   ],
   "devDependencies": {
-    "angularjs": "=1.3.16",
+    "angularjs": "#1.3.16",
```
jonman: Is `#` a valid thing?

jess0407: I think so, have changed `=` to `#` is because it is keep downloading 1.3.18, which broke the test.
with # it is now downloading 1.3.16
http://bower.io/docs/api/

gregbenner:
Tested and the login form works.

bower.json
gregbenner: I haven't seen the # I think the equals should be what we want ?http://stackoverflow.com/questions/22152232/bower-dependency-versions

nwhite89: what's the reason for using `#` with semver can't see an explanation of it online?

nwhite89: Yeah I think so have just installed & it seems to install see below :arrow_down: 

```sh
angular-elastic#ed898907fe vendor/angular-elastic
└── angular#1.3.18

angular-scroll#0.7.2 vendor/angular-scroll
└── angular#1.3.18

angular-aria#1.3.18 vendor/angular-aria
└── angular#1.3.18

angular-highlightjs#0.3.3 vendor/angular-highlightjs
├── angular#1.3.18
└── highlightjs#8.7.0

angular-animate#1.3.18 vendor/angular-animate
└── angular#1.3.18

angular-messages#1.3.18 vendor/angular-messages
└── angular#1.3.18

angular-touch#1.3.18 vendor/angular-touch
└── angular#1.3.18

angular-mocks#1.3.18 vendor/angular-mocks
└── angular#1.3.18

angular#1.3.18 vendor/angular

angularjs#1.3.16 vendor/angularjs

highlightjs#8.7.0 vendor/highlightjs

rbx_style_guide#bfb11b35c7 vendor/rbx_style_guide
├── angular-ui-router#0.2.15
└── angularjs#1.3.16

angular-strap#2.2.4 vendor/angular-strap
└── angular#1.3.18

angular-ui-router#0.2.15 vendor/angular-ui-router
└── angular#1.3.18
```

jess0407: @nwhite89 by using '=' or '#' to get the above result?

gregbenner: I think we don't need the equals '=' or the '#', it will implicitly get the specific version if major.minor.patch is specified like we have.

nwhite89: that is using `#` it should be `=` as far as I know

gregbenner: I can only find the '=' used in conjunction with > or < online...
I think if we write 1.2.3 instead of 1.2.x it will only ever get the exact of what we specified.
So "angular": "1.3.16" might be the best option?

jess0407: Strange, I was using `#` on Saturday and it downloaded angular.js 1.3.18 which cause all the tests to fail. Have changed to `#` to solve the problem. But have changed back to `=`, it seems to work now, so I'll keep it to `=`

nwhite89: https://docs.npmjs.com/misc/semver

```
Versions

A "version" is described by the v2.0.0 specification found at http://semver.org/.

A leading "=" or "v" character is stripped off and ignored.
```

PR #415, opened 2015-09-01
Fix radio group component

ohskylab:
TP: https://rockabox.tpondemand.com/entity/11567


gregbenner:
Ah looks good to moi

gregbenner:
:cat2:

PR #419, opened 2015-09-02
Add dynamic name to enable correct pairing of input to label tag

gregbenner:
Add dynamic name to enable correct pairing of input to label tag. Previously clicking an icon within a label would not highlight (select) the right input.

### DIFF
--- a/src/rb-check-control/rb-check-control-group.tpl.html
+++ b/src/rb-check-control/rb-check-control-group.tpl.html
@@ -34,7 +34,7 @@
     <div class="CheckControlGroup-body" ng-if="isDeviceChoice">
         <div ng-repeat="(key, value) in options">
             <rb-check-control
-                name="check-control-icon"
+                name="check-control-icon-{{::value.value}}"
                 label="{{::value.label}}"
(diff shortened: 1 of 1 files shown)


src/rb-check-control/rb-check-control-group.tpl.html
```diff
@@ -34,7 +34,7 @@
     <div class="CheckControlGroup-body" ng-if="isDeviceChoice">
         <div ng-repeat="(key, value) in options">
             <rb-check-control
-                name="check-control-icon"
+                name="check-control-icon-{{::value.value}}"
```
nwhite89: Not sure I'm following the Name correctly but should this not be `{{ ::value.label }}` ?

gregbenner: Naw should be as per the value, no spaces. Label would have spaces and special characters "Desktop / Mobile" vs "desktop-mobile". Needs a unique key for each compnent to link <input id="unique"> to <label for="unique">

nwhite89: oh I see yeah makes sense wasn't really following from the code

PR #421, opened 2015-09-03
Refactor rg-header to work with ui-router and state. Removes complexi…

gregbenner:
Refactor rg-header to work with ui-router and state. Removes complexity of transcluding attributes in favour of simple angular models. Update demo and tests.

### DIFF
--- a/src/rg-header/rg-header-item.tpl.html
+++ b/src/rg-header/rg-header-item.tpl.html
@@ -1,4 +1,18 @@
 <li class="RgHeader-navItem">
-    <a class="RgHeader-navItemInner" ng-transclude ng-class="{'is-active': isActive === 'true'}">
+    <a  class="RgHeader-navItemInner"
+        ng-if="!rgUiSref"
+        ui-sref-active="is-active"
+        ng-class="{'is-active' : isActive}"
+        ng-attr-target="{{target || undefined}}"
+        href="{{href}}"
+        ng-transclude>
+    </a>
+    <a  class="RgHeader-navItemInner"
+        ng-if="rgUiSref"
+        ui-sref="{{rgUiSref}}"
+        ui-sref-active="is-active"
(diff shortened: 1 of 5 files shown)


nwhite89:
@gregbenner no merge on this, you need to be able to set `is-active="true"` not just using `sref` this is a feature for Create and can't be removed can you please ensure that this is still possible

src/rg-header/rg-header-item.tpl.html
```diff
@@ -1,4 +1,7 @@
 <li class="RgHeader-navItem">
-    <a class="RgHeader-navItemInner" ng-transclude ng-class="{'is-active': isActive === 'true'}">
```
nwhite89: is active is a must needed.

gregbenner: For sure can be added back, won't conflict. Didn't see it used in the demo wasn't sure.

nwhite89: Can you add to the demo a version using active and one using `ui-sref-active` please cheers

gregbenner: Not sure how it functions as it never worked for us on Team Distribute.
I can add with ui-sref however it requires an added state and will show up as a component in the list on localhost:7000

nwhite89: not sure how what worked the `is-active`? you had an attribute you set as true 
```
<rb-header-item is-active="true"></rb-header-item>
```

nwhite89:
yeah go for it cheers :+1:

gregbenner:
Ready from my point of view for merge but please also check this if you want to merge https://github.com/rockabox/rbx_campaign_management/pull/268 should be done in parallel.

gregbenner:
@nwhite89 I ended up resolving with ng-if to avoid conflicts between href and lack of ui-sref, should be good to go.

PR #428, opened 2015-09-08
Selectable box component and tests

gregbenner:
A new component that has functionality like a radio group but looks like a box! rb-box-selector.
TP: https://rockabox.tpondemand.com/entity/11580

### DIFF
--- a/src/rb-box-selector/rb-box-selector.css
+++ b/src/rb-box-selector/rb-box-selector.css
@@ -1,0 +1,64 @@
+/** @define rbBoxSelector; use strict */
+
+/**
+ * The `.rbBoxSelector` component allows you to click and select a box
+ */
+
+:root {
+    --rbBoxSelector-background: var(--color-white-alabaster);
+    --rbBoxSelector-background-active: var(--color-green-accent);
+    --rbBoxSelector-border-color: var(--color-gray-mystic);
+    --rbBoxSelector-border-color-active: var(--color-green-base);
+    --rbBoxSelector-padding: 1em;
+    --rbBoxSelector-margin: 0 0.5em;
+    --rbBoxSelector-border-radius: 5px;
+    --rbBoxSelector-item-body: var(--color-gray-regent);
+    --rbBoxSelector-required-color: var(--color-orange-base);
+}
+
+/**
+ * 1. Comment
+ */
+
+.BoxSelector {
+    display: -webkit-flex;
+    display: -ms-flexbox;
+    display: flex;
+    -webkit-flex-wrap: no-wrap;
+    -ms-flex-wrap: no-wrap;
+    flex-wrap: no-wrap;
+    -webkit-justify-content: space-around;
+    -ms-flex-pack: distribute;
+    justify-content: space-around;
+}
+
+.BoxSelector-item {
+    border: 2px solid var(--rbBoxSelector-border-color);
+    background: var(--rbBoxSelector-background);
+    padding: var(--rbBoxSelector-padding);
+    border-radius: var(--rbBoxSelector-border-radius);
+    margin: var(--rbBoxSelector-margin);
+    flex-grow: 1;
+}
+
+.BoxSelector-item:hover {
+    cursor: pointer;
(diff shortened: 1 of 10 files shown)


src/rb-box-selector/rb-box-selector.css
```diff
@@ -0,0 +1,64 @@
...
+    border: 2px solid var(--rbBoxSelector-border);
+    background: var(--rbBoxSelector-background);
+    padding: var(--rbBoxSelector-padding);
+    border-radius: var(--rbBoxSelector-border-radius);
+    flex-grow: 1;
+    margin: var(--rbBoxSelector-margin);
+}
+
+.BoxSelector-item:hover {
+    cursor: pointer;
```
nwhite89: is there no colour change when hovering?

src/rb-box-selector/rb-box-selector.css
```diff
@@ -0,0 +1,64 @@
+/** @define rbBoxSelector; use strict */
+
+/**
+ * The `.rbBoxSelector` component allows you to click and select a box
+ */
+
+:root {
+    --rbBoxSelector-variable: var();
```
nwhite89: ???

src/rb-box-selector/rb-box-selector.css
```diff
@@ -0,0 +1,64 @@
...
+    --rbBoxSelector-background: #fcfcfc;
```
nwhite89: are none of the colours set within the current colour scheme? if they are these should be loaded using the top level variables

test/unit/rb-box-selector/rb-box-selector.spec.js
```diff
@@ -0,0 +1,78 @@
...
+            $scope.selections = selections;
+            $scope.myModel = $scope.selections[1];
+            compileTemplate('<rb-box-selector selections="selections" ng-model="myModel"></rb-box-selector>');
+        }));
+
+        describe('rb-box-selector', function () {
+
+            it('should have 3 options', function () {
+                expect(element.find('li').length === 3).toBe(true);
+                expect(element.find('li').length === 4).toBe(false);
```
nwhite89: not really sure this check is actually needed the one above is quite specific

gregbenner: The whole test, Thought it would be good to see it renders out the right amount of components?

src/rb-box-selector/rb-box-selector.css
gregbenner: Might be my bad I'm not sure where the variable colours would come from, I don't see any shared css within UI components?

src/rb-box-selector/rb-box-selector.css
gregbenner: I left what the rb-component-generator placed in, I can remove it, wasn't sure if it does something special.

src/rb-box-selector/rb-box-selector.css
gregbenner: I'll confirm with Rob.

src/rb-box-selector/rb-box-selector.css
nwhite89: Take a look at the style guide repo

src/rb-box-selector/rb-box-selector.css
nwhite89: no shouldn't be any way think is more of a placeholder for a variable As far as I am aware.

src/rb-box-selector/rb-box-selector.css
nwhite89: in fact I think it will be this file > https://github.com/rockabox/rbx_style_guide/blob/develop/src/assets/styles/variables.css

jess0407:
FYI: Have integrate this component into the updated new line item form. It works.

gregbenner:
Would be nice to merge at some point :)

gregbenner:
image

jess0407:
@gregbenner I think this is good and working as mentioned before. @nwhite89 still have any concerns on test?
If there is no concern on test anymore, I'll be happy to merge it.

PR #432, opened 2015-09-15
Change some layout to get rid of the double scroll bars on editor section of create.

gregbenner:
Change some layout to get rid of the double scroll bars on editor section of create.
TP: https://rockabox.tpondemand.com/entity/11534


gregbenner:
@jonman can you have a gander at your earliest?

jonman:
A tricky one. When the sidebar is longer than the screen, canvas will move on scroll whereas it should always be on the screen and only the sidebar should scroll.

gregbenner:
Ah think this might need some @ohskylab  magic... will close for now.

PR #433, opened 2015-09-15
Fix bug currency can be negtive

jess0407:
https://rockabox.tpondemand.com/entity/9470


jaylett:
There should probably be a unit test for this behaviour?

jess0407:
noted, will add

jess0407:
Test added :)

nwhite89:
Should there be a message to the user specifying that they can't add negative numbers? also this will need a squash

jess0407:
@nwhite89 have squashed. It will show user the message  'Please enter a valid monetary value.' which cover the case of a negative number.

nwhite89:
Oh yeah ok cool thanks @jess0407 will merge

PR #435, opened 2015-09-17
Select box error when no error present fix.

gregbenner:
Select box error when no error present fix.
TP: https://rockabox.tpondemand.com/entity/11388


jess0407:
Have tested on distribute, bug still persist.

jonman:
I think the problem is within `rbx_campaign_management` https://github.com/rockabox/rbx_campaign_management/blob/develop/src/states/ad-unit-form/new-ad-unit-form-ctrl.js#L20 where we override `brand` to be the `id` instead. We probably need to model `brand` separately.

```js
$this.brand = '';
..
$this.createAdUnit = function () {
    $this.adUnit.brand = $this.brand;
..
```
```html
<rb-select-control
    selected="vmAdUnitFormCtrl.brand"
  ..
```

PR #437, opened 2015-09-18
Add isActive and $state to sidebar items to ensure they are resolved active if they should be. Adjust test to mock $state module.

gregbenner:
Add isActive and $state to sidebar items to ensure they are resolved active if they should be. Adjust test to mock $state module.

TP: https://rockabox.tpondemand.com/entity/11772


jess0407:
Have tested on Distribute. LGTM

PR #439, opened 2015-09-21
Add back svg for warning component close since we removed the sprite sheet.

gregbenner:
Add back svg for warning component close since we removed the sprite sheet.

TP: https://rockabox.tpondemand.com/entity/11827


jonman:
Adding DNM label for now, left a comment on the TP ticket.

PR #441, opened 2015-09-28
Login area min width prevents resizing from errors

gregbenner:
Login area min width prevents resizing from errors.

TP: https://rockabox.tpondemand.com/entity/11742

### DIFF
--- a/src/rb-login-form/rb-login-form.css
+++ b/src/rb-login-form/rb-login-form.css
@@ -12,6 +12,11 @@
  * 1. Comment
  */
 
+/* Hack CSS */
+.hackLoginWidth {
+    min-width: 220px;
+}
+
 .LoginForm {
     color: var(--LoginForm-color);
 }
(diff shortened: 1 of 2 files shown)


src/rb-login-form/rb-login-form.css
```diff
@@ -12,6 +12,11 @@
  * 1. Comment
  */
 
+/* Hack CSS */
+.hack-width {
```
jonman: This is a permanent fix rather than a hack? I assume that the width will not change in the future.

gregbenner: @jonman  I think @ohskylab might have a better solution, The width selected is for the current longest error message. I can refactor as a more permanent solution if it's acceptable?

jonman: Oh I see. In that case could you make it confirm to his class name convention. `.hackWidth`?

gregbenner:
@jonman cool cool, updated className's

PR #443, opened 2015-10-01
Add more validation mechanism to text control

jess0407:
https://rockabox.tpondemand.com/entity/11855
- monetary value cannot be negative
- add max value attribute and validation
- add min value attribute and validation

test and merge https://github.com/rockabox/rbx_campaign_management/pull/343

### DIFF
--- a/src/rb-text-control/custom-validation-directive.js
+++ b/src/rb-text-control/custom-validation-directive.js
@@ -27,7 +27,8 @@ define([
                             maxDecimals = (splitSteps[1]) ? splitSteps[1].length : 0,
                             splitValue = (viewValue) ? viewValue.split('.') : [],
                             decimalsPlaces = (splitValue[1]) ? splitValue[1].length : 0,
-                            result = ctrl.$isEmpty(modelValue) || (decimalsPlaces <= maxDecimals);
+                            result = ctrl.$isEmpty(modelValue) ||
+                                     (decimalsPlaces <= maxDecimals) && (viewValue >= 0.01);
                         return result;
                     };
 
@@ -43,6 +44,18 @@ define([
                         result = ctrl.$isEmpty(modelValue) || (wholeNumberDigits <= 18);
                         return result;
                     };
+
+                    if (scope.maxValue) {
+                        ctrl.$validators.overMaxValue = function (modelValue, viewValue) {
+                            return ctrl.$isEmpty(modelValue) || (modelValue <= Number(scope.maxValue, 10));
+                        };
+                    }
+
+                    if (scope.minValue) {
+                        ctrl.$validators.underMinValue = function (modelValue, viewValue) {
+                            return ctrl.$isEmpty(modelValue) || (modelValue >= Number(scope.minValue, 10));
+                        };
+                    }
 
                 } if (scope.maxDigits) {
                     ctrl.$validators.maxDigits = function (modelValue, viewValue) {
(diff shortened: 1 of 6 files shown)


src/rb-text-control/rb-text-control-directive.js
```diff
@@ -50,7 +50,8 @@ define([
                 small: '@',
                 onChange: '&',
                 modelOptions: '@',
-                maxDigits: '@'
+                maxDigits: '@',
```
gregbenner: Why not use built in 'max' https://docs.angularjs.org/api/ng/input/input%5Bnumber%5D ?

jess0407: @gregbenner this pr is still under construction, purely to test on CM, and can be reviewed later.

gregbenner: Ok ignore it till you're ready :)

jess0407: Note the context of this component is rb-text-control, not native input, therefore build in min and max does not apply here.

gregbenner: I would just transclude the attribute onto the input and let the built in validator do the work.
Could still rename as you have, but let the native angular validation do the work, unless you need speciall functionality?

jess0407: Yep, the current implementation require less work, and it works. 
Note that angular produce validation attributes as something like myForm.input.$valid
But we need something like form[name].$error.exceedMaxValue and
form[name].$error.exceedMinValue to show and hide the relevant validation error message.
Can't see why should we reimplement this (it's not gonna improve much if any performance, and it's going to be more complicated and require more redundant code).

src/rb-text-control/custom-validation-directive.js
```diff
@@ -44,6 +45,26 @@ define([
...
+                        var result;
+                        if (!scope.maxValue) {
+                            return true;
+                        } else {
+                            result = (modelValue <= Number(scope.maxValue, 10));
+                        }
+                        return ctrl.$isEmpty(modelValue) || result;
+                    };
+
+                    ctrl.$validators.exceedMinValue = function (modelValue, viewValue) {
```
gregbenner: exceed min value?

jess0407: Apologies maybe my english is not that good.
But I believe 'exceed' make sense here just so as here below: 
http://www.programmingincpp.com/exceed-integer-limits.html
I am open to suggestions though, what is the a good word here to a native speaker?

gregbenner: No worries. exceed = be greater in number or size than (a quantity, number, or other measurable thing). "production costs have exceeded £60,000"
Would just keep it simple like:  overMax / underMin

src/rb-text-control/rb-text-control-directive.js
gregbenner: I would write the validations a bit simpler if it makes sense to you
<code>
if (scope.maxValue) {
	ctrl.$validators.exceedMaxValue = function (modelValue) {
		var result = (modelValue <= Number(scope.maxValue, 10)) || false;
		return result;
	}
} </code>
This way also the validator isn't attached and evaluated unless the scope property exists.

src/rb-text-control/custom-validation-directive.js
jess0407: Have updated using underMinValue and overMaxValue.

src/rb-text-control/rb-text-control-directive.js
jess0407: This is actually a good point, have refactored so as below:
screen shot 2015-10-05 at 17 45 13

PR #444, opened 2015-10-05
Use ng-show in select control for dynamic disabling.

jonman:
As part of a regression on rig create

TP: https://rockabox.tpondemand.com/entity/11912


nwhite89:
Looks good to me @gregbenner can you please check this fix with your work as it has introduced a large breaking change

gregbenner:
MoG

PR #449, opened 2015-10-19
Fixes a bug where side nav is not full width of its cell.

nwhite89:
When in a grid cell within the Overlay it should take the full width of it's parent setting as flex fixes this issue.

TP: https://rockabox.tpondemand.com/entity/12120


jim-dj:
Ready for merge to develop for QA

PR #451, opened 2015-10-20
Fix an issue where select control placement is shown in safari.

jonman:
`display: none` `option` will still show in safari. Changing this to `ng-if` will fix this.

(Tested for regression in rbx_campaign_management and rig_create)

TP: https://rockabox.tpondemand.com/entity/12125


nwhite89:
Looks good over to @rockabox/qa

rashmisagar26:
Ready for QA !

PR #463, opened 2015-12-02
Fix bug where switch components overlays the footer.

nwhite89:
Prevents a bug where the switch component would overlay the footer and other components due to z-index. 

screen shot 2015-12-02 at 10 34 56

TP: https://rockabox.tpondemand.com/entity/12499


jess0407:
@rockabox/qa over to you

PR #465, opened 2015-12-04
Bug fix minifed rb-text-control

nwhite89:
When minifying with html-minifier everything after `<textarea/>` is omitted as it is not valid HTML, this fixes the issue as a `<textarea>` can not be self closing by adding `</textarea>` also.

TP: https://rockabox.tpondemand.com/entity/12520


jim-dj:
Ready to merge for QA

PR #466, opened 2015-12-08
Reset fields look states when disabled.

nwhite89:
When a Text control is disabled reset the colour's back to the original as error's should no longer display.

TP: https://rockabox.tpondemand.com/entity/12519


jim-dj:
Can be merged for QA

jonman:
:+1: Looks fine

PR #467, opened 2015-12-10
Fix bug where IE11 not showing loggin screen

facundfortuny:
IE11 doesn't support min-height and needs an extra height to work. 

Added host to webpack  web server to be able to access via IP. 

TP: https://rockabox.tpondemand.com/entity/10806

### DIFF
--- a/src/rg-site/rg-site.css
+++ b/src/rg-site/rg-site.css
@@ -15,6 +15,7 @@
     display: flex;
     flex-direction: column;
     min-height: 100vh; /* 1 */
+    height: 100px;
 }
 
 .RgSite-middle {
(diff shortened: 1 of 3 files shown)


src/rg-site/rg-site.css
```diff
@@ -15,6 +15,7 @@
     display: flex;
     flex-direction: column;
     min-height: 100vh; /* 1 */
+    height:100px;
```
jonman: Looks fine but there should be a space `height: 100px;` between

facundfortuny:
@jonman I added the space :)

nwhite89:
Perhaps squash the commits ? @facundfortuny

jim-dj:
Can be merged for QA

PR #468, opened 2015-12-10
Freeze Angular Material version to rc6

darwinlopez:
No TP


nwhite89:
@rockabox/qa over to you :+1:

PR #475, opened 2016-01-11
Fix an issue with autofocus on login form

jonman:
TP: https://rockabox.tpondemand.com/entity/9663


nwhite89:
@rockabox/qa over to you

PR #493, opened 2016-06-14
Fix a clash between rb-overlay-modal and angular material md-menu

darwinlopez:
Sets overflow to null on destroy for overlay modals

JIRA: https://rockabox.atlassian.net/browse/RIG-1079


rashmisagar26:
Passed QA

